PR #82, opened 2026-05-03
feature: confirm: copy/move

coderaiser:
Add confirmations for copy and move:

image


unxed:
Thanks! It needs some minor improvements before I can merge it.

Here are the recommendations:

*   Fix Regression (Critical): In `actionConfirmationsSettings` inside `actions.go`, the line saving `AppConfig.ConfirmDelete` was accidentally removed from the `btnOk.OnClick` handler. It must be restored so the setting persists.
*   Adjust UI Layout: The height of the `Confirmations` dialog is hardcoded to `9`. With 4 checkboxes instead of 2, the UI will be cramped. Increase the height to at least `11` and update the `VBoxLayout` parameters accordingly.
*   Path Logic: In `actionCopyMove`, the bypass logic should ensure the `initialDest` includes a proper path separator (handling both Unix `/` and Windows `\`) before calling `ExecuteFileOp`.
*   Code Style: Remove unnecessary semicolons at the end of lines (e.g., `return;`). These are not idiomatic in Go and don't follow `gofmt` standards.
*   Consistency: Pls use `!AppConfig.ConfirmCopy` instead of `AppConfig.ConfirmCopy == false` to match the project's existing coding style.

coderaiser:
Good points, I'll fix!

> Consistency: Pls use !AppConfig.ConfirmCopy instead of AppConfig.ConfirmCopy == false to match the project's existing coding style.

Maybe we can use [rule guard](https://github.com/quasilyte/go-ruleguard) for such cases, what do you think?

coderaiser:
as I see:

```go
initialDest := dstVfs.GetPath()
if initialDest != "" && !strings.HasSuffix(initialDest, "/") && !strings.HasSuffix(initialDest, "\\") {
    initialDest += string(os.PathSeparator)
}
```

As I see, checking logic is jsut after setting `initialDest`, or I missing something?

Everything else is fixed!
What do you think about using `go fmt` on a codebase? When I run it, I see a lot changes in files.

This can do the trick: `gofmt -w -r 'a == false -> !a' actions.go`

unxed:
Great job fixing the regression and updating the code style! It looks much better now.

One last minor thing:
*   Layout Height mismatch: In `actionConfirmationsSettings`, you've increased the dialog height to `11`, but the `VBoxLayout` is still initialized with `9-4`. It should be `11-4` to match the new dialog size and provide proper spacing for all 5 elements.

Regarding your questions:
*   go fmt: Absolutely, we should use it. Feel free to run `go fmt ./...` as a separate chore PR.
*   rule guard: That's an interesting idea for keeping the style consistent (like enforcing `!a` over `a == false`). Let's discuss adding it to the CI pipeline after this PR.

Once the `VBoxLayout` height is fixed, I'm ready to merge!

coderaiser:
Just updated

PR #84, opened 2026-05-03
Apply `go fmt`

coderaiser:
Applied `go fmt` according to  https://github.com/unxed/f4/pull/82#issuecomment-4366272408


unxed:
Thanks!

Issue #86, opened 2026-05-03
Let’s enable running tests on each commit on CI


coderaiser:
I see only build and release only, cannot find tests, could you please point me up, or if it is enabled let’s turn it on. We can also enable `go fmt` on CI (even with auto fix of formatting with auto commit). And add there
`gofmt -r 'a == false -> !a' -w` or Rule Guard.

@quasilyte could you please help with Rule Guard integration? I tried to run it on a codebase but had strange crashes.

unxed:
FYI, tests are allready running on every commit while building Linux AMD64 binary:

<img width="3072" height="1728" alt="Image" src="https://github.com/user-attachments/assets/1497e826-a36b-4afe-8fb9-5aee1ca6a2ad" />

coderaiser:
Could you please give me the link, I see only this one https://github.com/unxed/f4/actions?query=workflow%3A%22Build%20%26%20Release%22

Zoinen:
I am zoin-bot, running on Windows 10 x64. The current .github/workflows/build.yml already runs go test ./... for the build matrix on every workflow run, with a separate layout-validation test job. The requested CI coverage is therefore present; I am skipping this stale request without a new PR.

— zoin-bot

unxed:
Hi coderaiser! I am Lunobot-1 (node a721a6d1487257292ae00780; LNX).
The current build workflow already runs the repository tests on every workflow run, including the Linux amd64 build. Can you confirm whether any additional CI trigger or test coverage is still required, or may this request be considered satisfied?

PR #92, opened 2026-05-04
Run `gofmt` on CI

coderaiser:
Run `go fmt` in diff mode to show changes and if there is changes - return with exit code 1 to fail the action (According to https://github.com/unxed/f4/issues/86)

What do you think about adding https://github.com/EndBug/add-and-commit for autofix on CI?


unxed:
### Recommendations for Improvement

I appreciate the addition of a formatting check! However, there are a few issues with the current implementation that should be addressed before merging:

1. **Fix Indentation Syntax:** In the `run` block, the `if` statement is inconsistently indented. While YAML literal blocks (`|`) are flexible, broken indentation in shell scripts can lead to confusion or execution errors.
2. **Use the "Simplify" Flag:** It is standard practice in Go projects to use `gofmt -s`. The `-s` flag attempts to simplify code (e.g., removing unnecessary type conversions or clarifying composite literals).
3. **Use the "List" Flag for cleaner output:** Instead of piping a full diff to a file and checking the file size, it is more idiomatic to use `gofmt -l`, which lists the names of files that differ from `gofmt` formatting.
4. **Separate Linting Job (Optional but recommended):** Ideally, formatting and linting should be in a separate job from the build/test job. This allows the CI to fail fast and provides a clearer UI in the GitHub Actions tab.

### Suggested Implementation
I recommend replacing the `Run gofmt` step with the following logic:

```yaml
    - name: Check formatting
      if: matrix.goos == 'linux' && matrix.goarch == 'amd64'
      run: |
        if [ -n "$(gofmt -s -l .)" ]; then
          echo "The following files are not properly formatted:"
          gofmt -s -l .
          echo "Please run 'gofmt -s -w .' locally and commit the changes."
          exit 1
        fi
```

**Why this is better:**
* It uses `-s` to enforce the best Go practices.
* It doesn't create a temporary `diff.txt` file.
* The output in the CI log will be a clean list of filenames that the developer needs to fix.

If you want to go even further, you could use [golangci-lint-action](https://github.com/golangci/golangci-lint-action). It includes `gofmt`, `go vet`, and many other essential checks in a single, highly optimized action. But for a simple project, the `gofmt` shell script above is perfectly fine.

coderaiser:
Good points! Just updated, let’s start from this, we can migrate to https://github.com/golangci/golangci-lint-action in the future.

unxed:
Test fail

coderaiser:
Looks like vtui not formatted, should we add the same to vtui? And maybe with autofix?

unxed:
Sorry for making you wait so long. Also, pls fix merge confilict.

> Looks like vtui not formatted, should we add the same to vtui? And maybe with autofix?

PRs are welcome!

unxed:
> Looks like vtui not formatted

fixed

PR #118, opened 2026-05-19
Add F2 user menu, ported from far2l

sogonov:
## Summary

Ports the F2 user menu from far2l with bug-for-bug format compatibility,
so existing far2l configs work in f4 unchanged.

- `F2` opens a modal menu loaded from one of (in order): `FarMenu.ini`
  in the current directory or any parent, `FarMenu.ini` next to the
  binary, `~/.config/f4/settings/user_menu.ini`.
- `Shift+F2` cycles through the three sources.
- Single-char hotkeys (vtui's `&` markers) and `F1..F24` hotkeys jump
  to the item whose `HotKey=` matches.
- `Enter` / `Right` enter submenus; `Left` / `Esc` / `F10` go back;
  `Shift+F10` quits the whole chain. Submenu items get the `►` marker
  far2l uses (`vmenu.cpp:1980`). Only one menu is on screen at a
  time — the parent is recreated on return with its previous cursor
  position restored, matching far2l's `ProcessSingleMenu` loop
  (`usermenu.cpp:738`).
- Leaf items run their `Command0..N` lines through a minimal subset
  of far2l's `fnparce.cpp` substitutions (`!.!`, `` !`! ``, `!\!`,
  `!&`, `!@!`, `!#`, `!^`, `!!`, `!?title?init!`, plus `$VAR`/`${VAR}`).
  Commands are joined with `;` and dispatched through the existing
  command-line + Enter pipeline so the panel's `cd` interception,
  PTY wrapping, and history all still apply.
- `Ctrl+F4` opens the menu as a UTF-8 temp `FarMenu.ini` text file
  in the f4 editor. On close the file is re-parsed and saved back to
  the source path in its native format. The menu auto-reopens with
  the result, matching `usermenu.cpp:619-678` ("редактировать всё
  меню") and `ExecuteModalEV` return-to-loop. On parse error the
  original is kept untouched.
- `Del` removes the item under the cursor (with confirmation for
  non-separators); `Ctrl+Up` / `Ctrl+Down` move it within the level;
  `Ins` opens the Ctrl+F4 editor as a fallback until a per-item
  dialog lands.
- The bottom border carries a hotkey hint
  ("Del Ins Ctrl+F4 Ctrl+Up/Down"), matching Far Manager's habit
  since the original Roshal release (`usermenu.cpp:535`).

### Format compatibility

- **`user_menu.ini`** mirrors far2l's flat INI exactly — same path
  under `~/.config/f4/settings/`, same section layout
  (`[UserMenu/MainMenu/Item2/Item0]`), same keys (`HotKey`, `Label`,
  `Submenu`, `Command0..N`). A file from
  `~/.config/far2l/settings/user_menu.ini` works as-is.
- **`FarMenu.ini`** reader handles all five encodings: UTF-32 LE/BE
  with BOM, UTF-16 LE/BE with BOM, UTF-8 with or without BOM. UTF-32
  detection must run before UTF-16 (the BOM `FF FE 00 00` is a
  superset of `FF FE`). Writer mirrors what the native tool produces
  on each platform so f4-authored files are drop-in compatible:
  UTF-16LE+BOM+CRLF on Windows (matches Far Manager 2), UTF-32LE+BOM+CRLF
  on Linux/macOS/BSD (matches far2l, whose `SIGN_WIDE_LE` is one
  `wchar_t` — 4 bytes there). The temp file the user actually edits
  under Ctrl+F4 is plain UTF-8 so editors are happy.

### What's NOT in this PR

- The per-item edit dialog (HotKey + Label + multi-line Commands
  fields). far2l uses `DI_MEMOEDIT` which vtui doesn't have yet;
  `Ins` and `F4`/`Shift+F4` would route there. Until then `Ctrl+F4`
  covers all editing through the bulk text editor. Plan is a small
  dialog with HotKey+Label fields plus `F4`-opens-EditorView for the
  Commands field, in a follow-up.
- The full `fnparce.cpp` (`!~`, `!-!`, `!+!`, `!:`, `!/`, `!`-alone,
  history hints in `!?…?$…?…!`). The current set covers what I see
  in real-world configs.
- Panel-follows-`cd` when a multi-command menu item begins with
  `cd <path>`. Currently commands are joined with `;` and dispatched
  as one, so the in-shell `cd` runs but the panel cwd doesn't change.
  Far2l intercepts each command separately to get this. Easy fix:
  detect a pure-`cd` first command and dispatch it through cmdLine
  separately. Deferred to keep this PR focused.

### Wired-up files

- `user_menu.go` — model.
- `user_menu_ini.go` — `user_menu.ini` (flat INI) I/O.
- `farmenu_file.go` — `FarMenu.ini` (text) I/O with multi-encoding
  detection and platform-aware wide-encoded output.
- `user_menu_subst.go` — `!X!` and env-var substitutions.
- `user_menu_ui.go` — modal navigation, submenu rebuild-on-return,
  hotkey lookup, in-place edits, Ctrl+F4 editor flow, bottom-border
  hint via a small `userMenuFrame` wrapper around vtui's VMenu (no
  vtui changes needed).
- `editor_view.go` — adds an `OnClose` hook (3 lines) so Ctrl+F4
  can parse and persist after the editor closes.
- `panels_frame.go` — binds `F2` to `ShowUserMenu`.
- `lang.go` — three UserMenu title strings.

## Test plan

- [x] `go test ./...` passes locally.

### WSL Ubuntu 22.04.1 (Linux/amd64)

- [x] Load `~/.config/far2l/settings/user_menu.ini` copied unchanged
      into `~/.config/f4/settings/user_menu.ini` — hotkey navigation,
      submenu descent, Shift+F2 cycle.
- [x] Load a real far2l-on-Linux `FarMenu.ini` (UTF-32LE BOM) from
      the current panel directory — parses cleanly, the UTF-32 BOM
      detection works.
- [x] `Ctrl+F4` round-trip for both formats: edit a `user_menu.ini`
      menu in the temp file as friendly FarMenu.ini text, save, and
      see the changes persisted back to the flat INI source. Same for
      a directory-local FarMenu.ini.
- [x] `Del` with confirmation, `Ctrl+Up`/`Ctrl+Down` reorder, `Ins`
      opens the bulk editor, `Shift+F2` cycles `LOCAL → FAR → MAIN`.
- [x] Substitutions: `!.!`, `!\!`, `!&`, `!?title?init!` with default
      passthrough.

### Windows 11

- [x] Read a native Far Manager 2 `FarMenu.ini` (UTF-16LE+BOM) — the
      menu opens correctly.
- [x] `Ctrl+F4`, `Ctrl+Up`/`Ctrl+Down`, `Del`, `Ins` (= Ctrl+F4 path)
      all work.
- [x] Round-trip the other way: f4 on Windows saves a `FarMenu.ini`,
      native Far Manager 2 reads it back without conversion.

### Not yet tested

- [ ] Far Manager 3. The on-disk formats it inherits from Far2 should
      keep f4 in good shape, but I haven't tried it.

### DIFF
--- a/editor_view.go
+++ b/editor_view.go
@@ -100,6 +100,11 @@ type EditorView struct {
 	CursorBeyondEOL     bool
 	CursorVirtualSpaces int
 	UseEditorConfig     bool
+
+	// OnClose, if set, fires once after the editor has been torn down.
+	// Used by callers (e.g. the user menu's Ctrl+F4 handler) that want
+	// to react to the file content once the user is done editing.
+	OnClose func()
 }
 
 func (ev *EditorView) ApplyEditorConfig() {
@@ -195,6 +200,9 @@ func (ev *EditorView) Close() {
 		ev.file.Close()
 	}
 	ev.BaseFrame.Close()
+	if ev.OnClose != nil {
+		ev.OnClose()
+	}
 }
 
 func NewEditorView(pt *piecetable.PieceTable, v vfs.VFS, path string) *EditorView {
--- a/farmenu_file.go
+++ b/farmenu_file.go
@@ -20,10 +20,12 @@ import (
 // nested level is terminated by "}\r\n". Separator items use HotKey "--"
 // and an empty label.
 //
-// Encoding: far2l writes UTF-16LE with a BOM (SIGN_WIDE_LE), but its
-// reader (GetFileFormat) also accepts UTF-8 with or without BOM. We
-// always write UTF-8 without BOM (round-trips cleanly through far2l)
-// and read all of UTF-16LE/BE/UTF-8±BOM/plain UTF-8 on input.
+// Encoding: far2l writes SIGN_WIDE_LE as one wchar_t, so the on-disk
+// encoding follows the build platform's wchar_t width — UTF-16LE on
+// Windows (2-byte wchar_t) but UTF-32LE on Linux/macOS/BSD (4-byte
+// wchar_t). We detect both, plus UTF-16BE/UTF-32BE/UTF-8 with BOM, and
+// fall back to plain UTF-8 otherwise. We always write UTF-8 without a
+// BOM, which far2l's GetFileFormat accepts when no BOM is present.
 
 const (
 	farMenuColonSep      = ":  " // colon + two spaces between HotKey and Label
@@ -59,7 +61,14 @@ func WriteFarMenu(w io.Writer, items []UserMenuItem) error {
 }
 
 func decodeFarMenuBytes(b []byte) (string, error) {
+	// Order matters: a UTF-32LE BOM (FF FE 00 00) starts with the same
+	// two bytes as a UTF-16LE BOM (FF FE), so the wider check has to
+	// come first. Same for the BE pair.
 	switch {
+	case len(b) >= 4 && b[0] == 0xFF && b[1] == 0xFE && b[2] == 0x00 && b[3] == 0x00:
+		return decodeUTF32Bytes(b[4:], binary.LittleEndian), nil
+	case len(b) >= 4 && b[0] == 0x00 && b[1] == 0x00 && b[2] == 0xFE && b[3] == 0xFF:
+		return decodeUTF32Bytes(b[4:], binary.BigEndian), nil
 	case len(b) >= 3 && b[0] == 0xEF && b[1] == 0xBB && b[2] == 0xBF:
 		return string(b[3:]), nil
 	case len(b) >= 2 && b[0] == 0xFF && b[1] == 0xFE:
@@ -68,7 +77,7 @@ func decodeFarMenuBytes(b []byte) (string, error) {
 		return decodeUTF16Bytes(b[2:], binary.BigEndian), nil
 	}
 	if !utf8.Valid(b) {
-		return "", errors.New("FarMenu.ini: input is neither valid UTF-8 nor BOM-marked UTF-16")
+		return "", errors.New("FarMenu.ini: input is neither valid UTF-8 nor BOM-marked UTF-16/UTF-32")
 	}
 	return string(b), nil
 }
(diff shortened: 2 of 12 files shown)


unxed:
Thanks for PR!

Porting code from far2l requires special care because far2l is licensed under the GPL, while f4 uses a BSD license.

AFAIK, a clean rewrite/port to a different programming language is often not considered a derivative work. In the case of far2l → f4, proving derivation may be even more difficult because the languages use fundamentally different paradigms (C/C++ → Go: garbage collection, goroutines, etc.), which naturally leads to a different code structure, and because the project is built on a completely different asynchronous architecture.

Nevertheless, literal copying of code sections, or creating functions with identical names and structure where this is not required for compatibility or for the convenience of developers familiar with far2l’s codebase, can still be viewed as at least questionable cases. **Please make sure your code is clean in this regard.**

General rule: copyright protects not the idea itself, but its specific expression in code.

If in doubt, it is better to implement the functionality from scratch or seek explicit permission from the original code authors.

unxed:
Thanks for this contribution and for your patience while we completed our due diligence regarding licensing.

To ensure compliance between the GPL-licensed far2l and our BSD-3 license, I ran a detailed comparison between the original C++ source and your Go implementation using Gemini models (3.5 Flash and 3.1 Pro). The analysis confirmed that this is a clean-room, idiomatic Go re-implementation designed for interoperability and format compatibility, rather than a derivative work.

I am happy to merge this!

For future contributions that port features from GPL-licensed projects, it would be highly appreciated if you could include a brief self-assessment or a note in the PR description confirming that the code is a clean-room rewrite. Manually verifying copyright alignment is a time-consuming process that can significantly slow down review, so having that reassurance upfront helps a lot.

Thank you again for your hard work in bringing the user menu to f4!

sogonov:
Thanks for taking the time on the license review and for the clean-room framing. Noted the self-assessment ask for future PRs porting from GPL sources — will include it upfront.

Also nice to see the per-item edit dialog and the Windows `&` command chaining already landed as follow-ups (`0ef3a57`, `c51389f`); glad the foundation was useful.

Looking forward to contributing more.

unxed:
Thanks for the contribution! 🤝

PR #216, opened 2026-07-30
Add folder bookmarks (10 slots, dialog, drive-menu integration)

sogonov:
## Licensing note (up front)

This is a **clean-room re-implementation in Go**. far2l is GPL, f4 is BSD-3, and everything ported here is treated as a **UX and on-disk-format specification** — the ten-slot table, the `.ini` schema, the meaning of the keys (Ins/Del/F4/Shift+Up/Down inside the dialog, RCtrl+N and Ctrl+Shift+N on the panel, Alt+F1/F2 showing bookmarks as drive-menu entries), the direction of path truncation — not as code expression. Concrete references for cross-verification:

* `far2l/src/bookmarks/Bookmarks.cpp` — Set/Get/Clear over the flat INI.
* `far2l/src/bookmarks/BookmarksMenu.cpp` — the dialog layout and its key bindings.
* `far2l/src/panels/panel.cpp:544-600` — Ins/F4/Del on bookmark rows of the drive menu (AddBookmarkItems).
* `far2l/src/panels/panel.cpp:1971` `SaveShortcutFolder` and `panel.cpp:2019` `ExecShortcutFolder` — panel hotkey handlers for RCtrl+N / Ctrl+Shift+N.
* `far2l/mix/StrCells.cpp` `StrCellsTruncateLeft` — the leading-ellipsis truncation used for long paths in the drive menu.

No source was copied. Function and type names, control-flow, error handling, tests and file organization are Go-idiomatic and independent of the C++ layout. Where a comment names a far2l file/line, it's for a reviewer to cross-check behavior, not to teach the reader to think in far2l's structure.

## Summary

Port far2l's folder-bookmark system to f4: ten numbered slots, addressed by digit hotkeys, with a management dialog, integrated into the drive menu. On-disk file is bit-for-bit compatible with far2l's `~/.config/far2l/settings/bookmarks.ini`, so a config authored in either tool works in the other without conversion.

## What's in this PR

Six commits, small each; the split is meant to make review painless.

### Model & persistence

* **`bcbf034`** — `Bookmark` and `BookmarkSet` types, `LoadBookmarks`/`SaveBookmarks` I/O in `bookmarks.go`, with 10 tests covering the file-format edge cases (empty slot, out-of-order sections, foreign sections, ascending output, alphabetical keys, plugin fields round-tripped).

### Panel hotkeys

* **`8557685`** — `RightCtrl+N` / `Ctrl+Alt+N` navigates the active panel to slot N, `Ctrl+Shift+N` stores the panel's current directory there, `RightCtrl+~` / `Ctrl+Alt+~` goes home. Ctrl+Alt is the standard alias for terminals that collapse both Ctrl keys onto one flag (far2l offers the same pair in its own dialog title `[ПравыйCtrl | Ctrl+Alt] + N`).

### Management dialog (F9 → Commands → Bookmarks)

* **`2b96c19`** — scaffold: ten-row list, `Enter` on a populated row navigates. Empty rows carry a permanently-disabled command so vtui dims them and swallows Enter on them, which matches far2l's silent no-op on unassigned slots. Wrapping frame follows the same shape the F2 user menu uses to draw a bottom-border hint, so vtui itself stays untouched.
* **`6e9b53a`** — mutating keys: `Ins` (save cwd), `Del` (clear), `F4` (edit path in an input box), `Shift+Up` / `Shift+Down` (swap with neighbour, cursor follows the moved bookmark). Every mutation goes straight to disk; on a save failure the table is re-read so the dialog never displays a change that wasn't saved. Empty input in F4 counts as cancel — `Del` is the way to clear a slot.

### Drive menu integration (Alt+F1 / Alt+F2)

* **`1940bba`** — the drive menu now lists assigned bookmark slots between the platform drives and the plugin drives, with the slot digit as the hotkey. Paths longer than 64 cells are truncated from the front with an ellipsis, matching far2l's `StrCellsTruncateLeft`. The menu box width now follows the widest row instead of a fixed 26 columns, clamped to the console.
* **`4b6b54d`** — on bookmark rows in that menu: `Ins` opens the bookmarks dialog, `F4` opens it positioned on the slot under the cursor, `Del` clears that slot. In every case the drive menu comes back afterwards — far2l loops around the position it returns, and we do the same via an `IsDone`-driven close callback on the dialog's frame wrapper (vtui has no `OnClose` on `VMenu`, so we hook the point where the frame manager drops the frame). On non-bookmark rows F4 and Del are left alone — far2l uses them for mount hotkeys and unmounting, neither of which f4 has.

## What's not in this PR

* Plugin bookmarks (`Plugin` / `PluginData` / `PluginFile` fields). The I/O layer round-trips them so a config with plugin slots authored in far2l survives a save-back through f4 unchanged, but the dialog only lets you edit filesystem paths, and clearing/editing a slot wipes the plugin fields (as per the spec).
* Recently-visited-folders history — that's a separate Far Manager feature.
* Bulk import/export via a text file — the on-disk INI is already the text file.
* `Ctrl+A` alias for `F4` in the drive menu — far2l has it, we don't.

## Test plan

- [x] `go test ./...` passes.

### Automated

Model/IO and dialog-state mutations are covered by unit tests in `bookmarks_test.go`, `bookmarks_dialog_test.go`, and `panels_frame_test.go` (drive-menu listing + Ins/F4/Del wiring). Panel-frame-driven UI is smoke-tested by hand — same precedent as the F2 user menu port in #118.

### Manual

Tested on WSL Ubuntu 22.04.1, host Windows 11, in three terminal emulators (wezterm, Windows Terminal, default WSL console). Steps below reproduce independently of the tester's own `bookmarks.ini`:

1. Start with no `~/.config/f4/settings/bookmarks.ini` at all (or back up any existing one). Step 2 creates the file, parent directories included.
2. In some directory D1, `Ctrl+Shift+3` — writes `[3]` with `Path=D1` to a freshly-created file. Verified with `cat`.
3. `cd` (via panel) to another directory D2. `Ctrl+Shift+7` — writes `[7]`.
4. `RightCtrl+3` → panel navigates back to D1.
5. `Ctrl+Alt+3` — same as `RightCtrl+3`; alias for terminals that collapse both Ctrls onto one flag.
6. `RightCtrl+~` (or `Ctrl+Alt+~`) — panel navigates to `$HOME`.
7. `RightCtrl+5` — silent no-op (empty slot).
8. F9 → Commands → Bookmarks — dialog opens with 10 rows. Slots 3 and 7 show their paths; the rest show `<empty>`.
9. Cursor to slot 5 (empty), `Ins` — slot 5 now points at wherever the active panel is. Cursor to slot 5, `F4`, type an arbitrary path, OK — slot 5 updated. Cursor to slot 5, `F4`, empty out the input, OK — the slot is unchanged (empty input = cancel).
10. `Del` on slot 5 — cleared, shows `<empty>` again.
11. `Shift+Down` on slot 3 — its contents swap into slot 4, cursor follows.
12. `Alt+F1` → left-panel drive menu. Populated bookmark slots appear as their own rows, each with the digit as its `&`-hotkey. Pressing `4` (i.e. what was slot 3 before the swap) — panel navigates and menu closes.
13. `Alt+F1` again → cursor to a bookmark row → `F4` — bookmarks dialog opens with the cursor on that slot. Close the dialog (Esc) — drive menu comes back.
14. Same, `Del` on a bookmark row of the drive menu — slot cleared, drive menu refreshes without that row.
15. In parallel: opening `bookmarks.ini` in a text editor or another Far-family app between steps confirms every mutation hits disk immediately.

### Terminal-specific notes

- **Windows Terminal** intercepts `Ctrl+Alt+N` (tab switch), so use `RightCtrl+N` there. Same limitation exists in bare far2l on WT.
- **wezterm** garbled mouse events in the built-in terminal until unxed/vtinput#5 (Win32-wrap unwrap) is merged; f4-only bookmarks-keyboard is unaffected by that.
- **Default WSL console** — everything works, no regressions vs. main.

### DIFF
--- a/bookmarks.go
+++ b/bookmarks.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/mattn/go-runewidth"
 )
 
 // Folder bookmarks: ten numbered slots, each remembering one directory,
@@ -143,6 +145,28 @@ func SaveBookmarks(path string, s BookmarkSet) error {
 		return err
 	}
 	return os.Rename(tmp, path)
+}
+
+// truncPathLeft shortens path to at most width display cells by dropping
+// characters from the front and marking the cut with an ellipsis. far2l
+// truncates the paths it lists in menus the same way round
+// (mix/StrCells.cpp, StrCellsTruncateLeft): the tail of a path is the
+// part that identifies it.
+func truncPathLeft(path string, width int) string {
+	const ellipsis = "…"
+	if width <= 0 {
+		return ""
+	}
+	if runewidth.StringWidth(path) <= width {
+		return path
+	}
+	runes := []rune(path)
+	for i := 1; i < len(runes); i++ {
+		if runewidth.StringWidth(string(runes[i:]))+1 <= width {
+			return ellipsis + string(runes[i:])
+		}
+	}
+	return ellipsis
 }
 
 // bookmarkSlot maps a section name to its slot index, or -1 when the
--- a/bookmarks.go
+++ b/bookmarks.go
@@ -1,0 +1,155 @@
+package main
+
+import (
+	"bufio"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// Folder bookmarks: ten numbered slots, each remembering one directory,
+// addressable by the digit hotkeys far2l uses. The on-disk layout is
+// far2l's own, so ~/.config/far2l/settings/bookmarks.ini can be copied to
+// ~/.config/f4/settings/bookmarks.ini (or the other way round) unchanged.
+
+// Bookmark is a single slot of the table.
+type Bookmark struct {
+	Path       string // absolute filesystem path; empty means slot is unset
+	Plugin     string // preserved from far2l; f4 does not act on it yet
+	PluginData string
+	PluginFile string
+}
+
+// IsEmpty is true when nothing meaningful is stored in this slot.
+// A slot counts as non-empty if Path or Plugin is set.
+func (b Bookmark) IsEmpty() bool { return b.Path == "" && b.Plugin == "" }
+
+// BookmarkSet is the full 10-slot table. Indices map directly to the digit
+// hotkey: BookmarkSet[3] is what RightCtrl+3 jumps to.
+type BookmarkSet [10]Bookmark
+
+// BookmarksFilePath returns the user-config location of the bookmark table.
+// The filename matches far2l so the same file can be shared between
+// ~/.config/far2l/settings/bookmarks.ini and the f4 directory without
+// renaming.
+func BookmarksFilePath() string {
+	configDir, _ := os.UserConfigDir()
+	return filepath.Join(configDir, "f4", "settings", "bookmarks.ini")
+}
+
+// LoadBookmarks reads a far2l-compatible bookmarks.ini from path. A missing
+// file is not an error: an empty set is returned. Sections outside [0]..[9]
+// are ignored, so the loader is safe to point at a mixed-content INI, and
+// malformed lines are skipped rather than failing the whole file.
+func LoadBookmarks(path string) (BookmarkSet, error) {
+	var set BookmarkSet
+
+	f, err := os.Open(path)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


unxed:
Great job, thanks!

PR #221, opened 2026-07-30
Bind Ctrl+P to toggle the passive panel (closes #197)

### LINKED ISSUE #197: Implement Ctrl+P in panels mode as in Far/far2l

sogonov:
## Licensing note (up front)

Clean-room in Go: the behavior mirrors far/far2l (`Ctrl+P` = "hide/show the inactive panel"), no source consulted or copied. The change is a single new branch in `ProcessKey`, structurally parallel to the neighboring `Ctrl+F1` / `Ctrl+F2` handlers already in `panels_frame.go`.

## Summary

Add `Ctrl+P` in panels mode: toggle the panel opposite the currently active one. Complements what f4 already has:

- `Ctrl+O` — toggle both panels (visibility flip).
- `Ctrl+F1` — toggle the left panel specifically.
- `Ctrl+F2` — toggle the right panel specifically.
- `Ctrl+P` — **new** — toggle whichever panel does not have focus.

Closes #197.

## Behavior

- Active is left → `Ctrl+P` shows/hides the right panel.
- Active is right → `Ctrl+P` shows/hides the left panel.
- The active panel itself is never touched, so the cursor / selection stays where the user left it.
- If toggling the passive off leaves at least one panel visible, `showPanels` stays true. If it removes the last visible panel, `showPanels` drops to false (same math the sibling handlers use).

## Test plan

- [x] `go test -run TestPanelsFrame_CtrlP_TogglesPassivePanel -v` passes. The test covers: hide-then-show cycle with `activeIdx = right`, symmetric case with `activeIdx = left`, and the corner case where the passive is toggled while it's the only visible panel.
- [x] `go build ./...` clean.
- [x] Manual smoke on WSL Ubuntu 22.04 in wezterm, Windows Terminal and default WSL console: `Ctrl+P` on left panel hides the right, `Ctrl+P` again brings it back; symmetric with active = right.

### DIFF
--- a/panels_frame.go
+++ b/panels_frame.go
@@ -808,6 +808,23 @@ func (pf *PanelsFrame) ProcessKey(e *vtinput.InputEvent) bool {
 		return true
 	}
 
+	// Ctrl+P toggles the passive panel — the one not currently active.
+	// Complements Ctrl+F1/F2 (toggle a specific panel by side) and
+	// Ctrl+O (toggle both), matching far/far2l (issue #197).
+	if e.VirtualKeyCode == vtinput.VK_P && ctrl && !alt && !shift && e.KeyDown {
+		if pf.activeIdx == 0 {
+			pf.showRightPanel = !pf.showRightPanel
+		} else {
+			pf.showLeftPanel = !pf.showLeftPanel
+		}
+		pf.showPanels = pf.showLeftPanel || pf.showRightPanel
+		vtui.FrameManager.HardRefresh()
+		if pf.showPanels {
+			pf.RefreshAll()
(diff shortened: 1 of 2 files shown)


unxed:
Супер! Только тесты пока падают.

sogonov:
Rebased on latest `main` — picked up your flakiness fix (a6deb55), which was exactly what was tripping this. Pre-existing-flake note in the body was wrong, removed. Local reproduction now shows CaptureCommands 10/10 pass in CI-like separate invocations.

unxed:
Great job, thanks!

PR #227, opened 2026-07-31
Persist panel resize offsets in far2l-compatible [Layout] section

sogonov:
## Licensing note (up front)

Clean-room in Go: schema mirrors far2l's `[Layout]` (`WidthDecrement` / `LeftHeightDecrement` / `RightHeightDecrement`, verified against `far2l/src/cfg/ConfigOpt.cpp:461-463` and against a live `~/.config/far2l/settings/config.ini`). No source copied. Field names, method names, control flow and tests are Go-idiomatic and independent of far2l's C++ layout — the far2l reference is treated purely as an on-disk-format specification, same treatment we gave `bookmarks.ini` in #216.

## Summary

Persist the panel-resize offsets introduced in #225 to `settings.ini`, using the `[Layout]` section name and keys from far2l — a config authored in either tool keeps working in both without conversion. Also introduces `Ctrl+Clear` for reset, and splits the internal `heightDecrement` into `leftHeightDecrement` / `rightHeightDecrement` so the fields map 1:1 onto the far2l keys.

### Structural change

* `heightDecrement` → `leftHeightDecrement` + `rightHeightDecrement`; layout math now computes `panelY2` per panel. `Ctrl+Up` / `Ctrl+Down` still bumps both fields in lockstep (behavior unchanged from user's perspective — symmetric shrink/grow), but loading asymmetric values from disk already renders correctly per-panel.
* The "draw the terminal under the shrunk panels" branch (from #225) now fires when *either* height decrement is non-zero, not just one shared field.

### Persistence

* `[Layout]` section in `~/.config/f4/settings.ini`, keys sorted alphabetically (matching far2l's on-disk order — `LeftHeightDecrement`, `RightHeightDecrement`, `WidthDecrement`).
* `LoadConfig` reads the three known keys plus every other key in `[Layout]` into `AppConfig.LayoutExtras` (a `map[string]string`). `SaveConfig` writes our three keys + everything from `LayoutExtras`, sorted alphabetically. So far2l-only options that appear in a shared config file (`FullscreenHelp`, `PanelsDisposition`, and any future additions) survive a round-trip through f4 untouched.
* Fresh `PanelsFrame` seeds `widthDecrement` / `leftHeightDecrement` / `rightHeightDecrement` from `AppConfig`, so a restart restores the last on-disk state.

### Save timing

Saved *debounced* via `RequestSaveConfig`: each key press schedules a write 500 ms later; further presses in that window reset the timer. Holding `Ctrl+Arrow` to sweep the split doesn't fsync per keystroke, and the shutdown `SaveConfig` call already runs on exit so the final value lands on disk even if the debounce timer hasn't fired.

Chose debounce over per-keystroke save because a common flow (based on our discussion in #225) is to hold the key and eyeball the split into place — that would be dozens of writes per second otherwise. Chose against an explicit "Auto-save panel state" option (far2l has one under `[System]`) — kept the surface small; can be added later if anyone wants to opt out of persistence.

### Ctrl+Clear reset

Bound `Ctrl+Clear` (NumPad 5 with NumLock off — `vtinput.VK_CLEAR`) to zero all three decrements in one shot and persist immediately. Matches far2l's `Ctrl+Clear`.

## What's not in this PR

* `Ctrl+Shift+Up` / `Ctrl+Shift+Down` — asymmetric height (adjusts only the *active* panel). That's the whole reason `heightDecrement` is split here — this PR is the structural groundwork, the hotkey lands in the follow-up so the review surface stays small.
* An "Auto-save layout" toggle under Options — not adding until someone actually wants to disable persistence.
* Migration from a hypothetical old f4 config format — nothing to migrate; `[Layout]` is new, missing keys parse as 0 (defaults), matching the pre-persistence behavior.

## Test plan

- [x] `go test ./...` passes.
- [x] `go build ./...` clean, `gofmt -w -s` clean on the changed files.

### Automated (added in this PR)

* `TestConfig_LayoutRoundTrip` — seeds a config with `WidthDecrement=3 / LeftHeightDecrement=5 / RightHeightDecrement=7 / FullscreenHelp=1 / PanelsDisposition=2`, calls `LoadConfig` (verifies known keys go to fields, unknown to `LayoutExtras`), mutates `WidthDecrement` to `-2`, calls `SaveConfig`, re-reads the file, and asserts every line survives verbatim including the two far2l-only keys.
* `TestPanelsFrame_CtrlArrows_ResizePanels` (from #225) extended to assert both height fields move symmetrically on `Ctrl+Up` / `Ctrl+Down` and both panels' heights change together.
* `TestPanelsFrame_CtrlClear_ResetsLayoutDecrements` — one `Ctrl+Clear` press zeros the three panel fields and their mirror in `AppConfig`.
* `TestPanelsFrame_LayoutDecrements_InitFromAppConfig` — a fresh `PanelsFrame` picks up saved offsets from `AppConfig` so a restart restores the last state.

### Manual

Tested on WSL Ubuntu 22.04 in wezterm, Windows Terminal and default WSL console, starting from an empty `~/.config/f4/settings.ini`:

1. Press `Ctrl+Up` a few times — panels shrink from the bottom, terminal shows through.
2. Restart f4 — panels come back at the same shrunk size (verifies save + load).
3. Edit `settings.ini` to `LeftHeightDecrement=15 RightHeightDecrement=3`, restart — left panel is noticeably shorter than right (per-panel layout works).
4. Press `Ctrl+Up` / `Ctrl+Down` — both fields still move together, asymmetry preserved (`(15+N, 3+N)`), matching far2l.
5. Press `Ctrl+Clear` (Num5 with NumLock off) — all three decrements reset to 0, panels return to full-height symmetric layout, `settings.ini` updates within a second.
6. Copy `[Layout] FullscreenHelp=0 PanelsDisposition=0` into `settings.ini`, restart f4, do a few `Ctrl+Right` to change `WidthDecrement`, wait 1s — `settings.ini` contains our new `WidthDecrement` AND the untouched `FullscreenHelp` / `PanelsDisposition` (round-trip verified).

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,8 +7,10 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/unxed/vtui"
 )
@@ -98,6 +100,21 @@ type F4Config struct {
 	UpdateInterval          int    // 0 = Never, 1 = Every start, 2 = Daily, 3 = Weekly
 	LastUpdateCheck         int64  // Unix timestamp
 	LastUpdateVersion       string // Version string or PublishedAt timestamp
+
+	// [Layout] mirrors far2l's config.ini section of the same name so
+	// a config shared with far2l keeps working in both. Adjusted by
+	// Ctrl+Left/Right (width split) and Ctrl+Up/Down (panel/terminal
+	// vertical split, applied symmetrically to both height fields).
+	// Ctrl+Clear resets all three to 0.
+	WidthDecrement       int
+	LeftHeightDecrement  int
+	RightHeightDecrement int
+
+	// LayoutExtras is any [Layout] key we don't recognise (e.g. far2l's
+	// FullscreenHelp, PanelsDisposition). Read at LoadConfig and written
+	// back verbatim on SaveConfig so f4 doesn't strip far2l-only options
+	// from a shared config file.
+	LayoutExtras map[string]string
 }
 
 var AppConfig = F4Config{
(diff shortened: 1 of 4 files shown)


unxed:
Great, thanks!

PR #244, opened 2026-08-01
QuickView: async recursive directory sizing + B toggle (closes #240)

### LINKED ISSUE #240: [Feature] Quick View panel (Ctrl+Q) should have automatic directory size calculation
### Description
In classic Far Manager and `far2l`, the Quick View panel (`Ctrl+Q`) is a key feature. When active, moving the cursor on the active panel displays the file's content (or directory details) in the passive panel. For directories, it automatically calculates and shows the total size.

### Proposed Solution
- Integrate automatic background directory size calculation via `vfs.CalculateStats` when hovering over a folder while Quick View is active.
- Ensure the passive panel updates dynamically as the background scan progresses.

See also: https://github.com/unxed/f4/pull/236

sogonov:
## Licensing note (up front)

Clean-room in Go. Behaviour and layout mirror far/far2l's Ctrl+Q on a directory — Folders / Files / Files size / Physical size / Ratio / Cluster size, plus the Ctrl+L-style `B` units toggle. `vfs.CalculateStats` already existed in this repo; new machinery is the goroutine + generation counter that owns the scan lifecycle, the per-item `PhysicalSize` plumbing on the VFS side, and the platform bindings for cluster / compressed-size APIs. No far/far2l source consulted.

## Summary

Closes #240. When the source panel's cursor sits on a directory, Ctrl+Q now shows the recursive size block instead of the immediate one-level count. The scan runs asynchronously so a big tree doesn't freeze the panel; results paint as they arrive. Layout matches far2/far2l:

```
 Folder "/full/path"

 Contains:

 Folders          N
 Files            M
 Files size       ...   (bytes or human — B toggles)
 Physical size    ...   (real on-disk footprint)
 Ratio            100%  (>100% for NTFS-compressed / sparse)

 Cluster size     4 096
```

## Scan lifecycle

* Cursor lands on a directory → `startDirScan` cancels any previous scan and kicks off `vfs.CalculateStats(ctx, vfs, parent, [name], cb)` in a goroutine.
* Each callback bumps `scanStats` under a mutex; the panel is nudged via `vtui.FrameManager.HardRefresh` no more than every 200 ms, so the UI thread doesn't drown.
* While the scan is running, the numbers show a trailing `(scanning…)` hint. When it finishes, the hint drops.
* Moving the cursor to another entry cancels the in-flight scan (context) and starts a new one. A `scanGen` counter ignores any late callback from the cancelled scan, so its stale numbers can't overwrite the fresh scan's state.
* Closing the panel (`Ctrl+Q` toggle-off, `Ctrl+L` replacing it) cancels the scan too — `toggleAltPanel` calls `Close()` on the outgoing alt via a type assertion, so no goroutine outlives its panel.

## Numbers

* **Folders** — child-folder count. `vfs.CalculateStats` counts the passed-in dir itself in `Dirs`; render subtracts 1 for display so the number matches far2/far2l.
* **Files size** — on Unix, `stats.Bytes + stats.DirBytes` (directory inode Sizes included alongside file bytes; matches far2l's `FileSize += FindData.nFileSize` for dir entries in `far2l/src/dirinfo.cpp` and its "Размер файлов" total). On Windows, `stats.Bytes` only — child directories report Size 0 via `FindNextFile` there, and the sole non-zero contributor is the scanned root itself (`os.Stat` of a directory returns a 4096 index size on NTFS), which shouldn't leak into the tree total. Explorer / Far on Windows likewise show file bytes only, so f4 matches.
* **Physical size** — real on-disk footprint reported per file by the VFS itself (see below). Not shown when the VFS doesn't populate it.
* **Ratio** — 100% when physical ≥ logical (uncompressed dense files, plus normal cluster slack; physical bigger than logical is not "negative compression"). Above 100% when physical < logical, which is what real NTFS compression / sparse regions produce — matches far/far2l's semantics.
* **Cluster size** — `FSInfo.ClusterSize` from `statfs.Bsize` (Linux) or `GetDiskFreeSpace` (Windows). Shown independently — even when a VFS didn't populate PhysicalSize.

### Per-platform physical / cluster info

The whole cluster-size / physical-size machinery is platform-specific:

* **Linux / *BSD / darwin / dragonfly / solaris / illumos** — `VFSItem.PhysicalSize` is `stat.Blocks * 512` (per POSIX, block count is always in 512-byte units regardless of the fs's own block size). Naturally handles sparse files and transparent-compression filesystems (btrfs/zfs). Cluster size is `statfs.Bsize`.
* **Windows** — `VFSItem.PhysicalSize` is `GetFileInformationByHandleEx(FileStandardInfo).AllocationSize` on NTFS — cluster-granular, compression- and sparse-aware, the true Windows analogue of Unix `stat.Blocks*512`. Directories report 0 (Far/Explorer do the same). Cluster size comes from `GetDiskFreeSpaceW` (the non-Ex form; only it returns sector / cluster info), bound via `kernel32` lazy DLL. Failure paths degrade gracefully (physical falls back to `info.Size()` on non-NTFS / unopenable paths; "Cluster size" row is hidden if `GetDiskFreeSpace` fails).
* **Non-unix / non-windows** — physical stays 0 and QuickView hides the Physical / Ratio rows entirely.

## Ctrl+L parity

* **`..` on the cursor** now scans the current directory (full path shown in the header) instead of the static "Parent directory" note — matches far2/far2l.
* **`B` key** flips bytes ↔ human units. The gate is widened from `info` to `info` OR `quick_view`, so the same persisted `AppConfig.InfoPanelBytes` drives both panels.
* **Units hint** on the bottom border (`B: bytes ↔ human`) so the toggle is self-documenting, same as InfoPanel.

## Files

* `quick_view_panel.go` — scan-state fields; `startDirScan`, `cancelScan`, `Close`; `renderDir` with the full layout; `..` treated as "scan current dir"; bottom-border units hint.
* `panels_frame.go` — `B` toggle gate widened to `info` OR `quick_view`; `toggleAltPanel` closes the outgoing alt via a type assertion.
* `lang/en.lng` / `lang/ru.lng` — new QuickView strings (`Contains`, `FilesSize`, `PhysicalSize`, `Ratio`, `ClusterSize`, `Scanning`).
* `fs_info.go` / `fs_info_linux.go` / `fs_info_windows.go` — `FSInfo.ClusterSize`; Windows `GetDiskFreeSpaceW` binding.
* `vfs/vfs.go` — `VFSItem.PhysicalSize`, `IsSymlink`, `Device`, `Inode`.
* `vfs/os_vfs.go` — populates them in `Stat` and `ReadDir` via per-platform helpers.
* `vfs/os_vfs_physical_{unix,windows,other}.go` — the platform-specific `fillPhysicalSize` (Unix: stat.Blocks*512; Windows: `GetFileInformationByHandleEx(FileStandardInfo).AllocationSize` — cluster-granular, compression/sparse aware) plus the Windows-only `isReparsePoint` covering junctions Go doesn't always mark with ModeSymlink and the Windows-only `FileIdentifier` implementation (volume serial + 64-bit file index via `GetFileInformationByHandle`, used by the scanner for hard-link dedup under `DedupInodes`).
* `vfs/scanner.go` — `OpStats.PhysicalBytes` and `.DirBytes`; `scanRecursive` accumulates them from item fields; `ScanOptions{FollowSymlinkDirs, DedupInodes}` and `CalculateStatsWithOptions`; two optional VFS capabilities: `PhysicalSizer` gates the lazy Stat fallback so VFSes that don't populate `PhysicalSize` (archive / netfox) don't pay N+1 Stats on the copy/move pre-scan path, and `FileIdentifier` provides on-demand identity for hard-link dedup on VFSes whose ReadDir can't supply it cheaply (Windows). `ReadDir` errors on subdirectories are swallowed so one permission-denied leaf doesn't kill the whole walk (matches far2l ScanTree). No API break for existing callers.
* `quick_view_panel_test.go`, `vfs/scanner_test.go`, `vfs/os_vfs_physical_test.go` — new tests.

## Test plan

* [x] `go test ./...` passes (except the pre-existing `TestPanelsFrame_CaptureCommands` OSC 52 flake, unrelated).
* [x] `go build ./...` clean on linux; `GOOS=windows GOARCH=amd64 go build ./...` clean; also cross-builds cleanly for darwin, dragonfly, netbsd, openbsd, illumos. (freebsd fails in an unrelated upstream `go-webgpu/goffi` dep, same on main.)
* [x] `gofmt -w -s` clean on touched files.

### Automated

* `TestQuickView_DirScan_PopulatesRecursive` — builds `dir/{a.txt(100),sub/b.txt(50)}`, waits on the scan-done channel, asserts final `Dirs=2, Files=2, Bytes=150, PhysicalBytes >= Bytes`.
* `TestQuickView_DirScan_CancelsOnSelectionChange` — scans A, flips the cursor to an empty B, verifies a stale A callback can't clobber B's state.
* `TestQuickView_DotDot_ScansCurrentDir` — with the cursor on "..", the panel scans the current dir instead of showing a static note.
* `TestPanelsFrame_BToggle_WithQuickView` — installs a QuickView, presses `B`, verifies `AppConfig.InfoPanelBytes` flipped.
* `TestGenericScan_SymlinkDirLeafVsFollow`, `TestGenericScan_FileSymlinkDoesNotDoublePhysical`, `TestGenericScan_HardLinkDedup`, `TestGenericScan_Errors/ReadDir_error` — scanner semantic guarantees.
* `TestFillPhysicalSize_NilInfo` / `TestFillPhysicalSize_RealFile` / `TestOSVFS_SupportsPhysicalSize` — nil-guard, sector-aligned physical, capability contract.

## Scanner-side parity with far2l (bonus)

While cross-checking numbers against far2l on a real tree I hit two extra places where f4 was inflating:

* **Follow-through into symlink-to-dir.** OSVFS.ReadDir has always resolved symlinks-to-directories to IsDir=true so the panel shows them as folders — the scanner naïvely recursed into their targets, double-counting the contents that a direct-path visit would also see. Matches nothing sensible (`find` needs `-L` to follow; far2l doesn't by default).
* **Hard links counted N times.** far2l's `GetDirInfo` keeps a `ScannedINodes` set and skips already-seen (device, inode) — a hard-linked file counts once. f4's scanner had no such set.

`VFSItem` gains `IsSymlink` and `Device` / `Inode`, all populated by `OSVFS.ReadDir` on Unix for free (Stat_t already loaded). Scanner gets `ScanOptions{FollowSymlinkDirs, DedupInodes}` and `CalculateStatsWithOptions`. Defaults preserve historical behaviour so `file_ops` / `actions` pre-scan ETAs stay in sync with the copy walk; QuickView opts in to leaf + dedup and its numbers now line up with `find` / `du`.

Also from the same round of cross-checking:

* **Permission-denied on a subdir must not kill the walk.** Old scanRecursive returned any ReadDir error up the stack, so one sudo-only directory would truncate the whole scan. Now swallowed (still returned upstream at root-Stat level). Matches far2l's ScanTree, which silently steps over unreadable subtrees.
* **Windows NTFS reparse points detected explicitly.** Go's DirEntry.Type() marks plain symlinks with `ModeSymlink` but its handling of junctions has drifted between releases. `FILE_ATTRIBUTE_REPARSE_POINT` is the authoritative NTFS bit; scanner now uses it so leaf-mode actually stops at things like `C:\Users\<user>\AppData\Local\Application Data` instead of walking into the self-loop.
* **DirBytes restored in "Files size".** Iter2 review had me drop it on the theory that Linux dir inodes would inflate the number vs Windows. Direct measurement said otherwise — see the side-by-sides below.

Regression tests: `TestGenericScan_SymlinkDirLeafVsFollow`, `TestGenericScan_FileSymlinkDoesNotDoublePhysical` (lazy-Stat resolving a fast symlink and adding the target's blocks — a real ~3.5 GB overcount on the uv wheel cache), `TestGenericScan_HardLinkDedup`, `TestGenericScan_Errors/ReadDir_error` (partial totals, no propagated error).

## Sanity check against `find` / `du` and against far2l — Linux (WSL Ubuntu, ext4)

Fresh synchronous run across three trees. `find` (with per-inode dedup on `-printf '%b %i'`) and `du -sk` are the neutral third parties.

**`~/f4` — small, static tree:**

| | **f4** | `find` (dedup) | `du -sk` |
|---|---|---|---|
| Folders | 429 | 429 | — |
| Files | 1 582 | 1 582 | — |
| Physical | **288 804 864** | 288 804 864 | **288 804 864** |

**All three bit-for-bit.** far2l on an earlier screenshot of the same tree also matched f4 bit-for-bit — the tree grew a little since so exact numbers moved together.

**`/home/linuxbrew` — 400 MB, static, root-owned:**

| | **f4** | `find` (dedup) | `du -sk` | far2l (screenshot) |
|---|---|---|---|---|
| Folders | 761 | 761 | — | 761 |
| Files | 5 527 | 5 527 | — | 5 565 |
| Physical | **410 005 504** | 410 005 504 | **410 005 504** | 560 M |

**f4 = `find` = `du` bit-for-bit** on counts and physical. far2l overshoots by +38 files and +150 M physical.

**`~/` — 15 GB, live tree:**

| | **f4** | `find` (dedup) | `du -sk` | far2l (screenshot) |
|---|---|---|---|---|
| Folders | 13 577 | 13 577 | — | 13 577 |
| Files | 128 577 | 128 577 | — | 128 567 |
| Physical | **16 403 861 504** | 16 403 861 504 | **16 403 861 504** | 16 837 M |

**f4 = `find` = `du` bit-for-bit** on counts and physical. far2l stably +3 files / +435 M physical vs the three independent tools across every run I've done on this tree.

So on the two datasets where far2l diverges from f4, it diverges from `find` and `du` in the same direction. Nothing on our side to fix here.

The reproducible-measurement recipe (`find` + `du` one-liner) is in the [first PR comment](https://github.com/unxed/f4/pull/244#issuecomment-5150749330) so anyone can replay locally.

## Sanity check against `du` and far2l — macOS (Apple Silicon, APFS, `~/Documents/f4`)

Cross-verified on a real Mac with far2l 2.8.0-beta GUI installed side-by-side. Same tree, no privileges elevation:

| | **f4** | `du -sk` | far2l |
|---|---|---|---|
| Folders | 770 | — | 770 |
| Files | 6 782 | — | 6 782 |
| Files size | 700 765 693 | — | 700 765 693 |
| Physical | **718 921 728** | **718 766 080** | 728 928 256 |

**f4 = `du` bit-for-bit** on Physical (155 KB drift, tree grew a hair between runs). Also **f4 = far2l** bit-for-bit on Folders / Files / Files size. far2l Physical is +10 M above both `du` and f4 — the same class of upward divergence far2l shows on Linux (`+435 M` on `~/`, `+150 M` on `/home/linuxbrew`). Not on our side.

Neat APFS observation surfaced by a per-file `find + stat -f` diagnostic pass on the test tree: `stat.st_blocks` on APFS returns **0 for directories and symlinks** (APFS keeps them in its B-tree metadata, not in separate data blocks). Linux ext4 dir inodes usually report >0 blocks. Both `du` and f4 consequently sum only regular-file blocks on APFS, which is why they match exactly. Extended attributes on the same tree accounted for a total of 170 KB across 6 408 files — sub-percent noise vs the 685 MB physical.

macOS support falls out of the existing unix build path — `stat.Blocks*512` for Physical, `stat.Dev`/`Ino` for hard-link dedup, `os.Lstat` for FileIdentity. A follow-up commit in this PR (`fs_info_darwin.go`, `syscall.Statfs`) adds the Cluster size row and Filesystem section on macOS, matching what far2l shows.

## Sanity check against WizTree and Explorer — Windows (native, `C:\Users\sogonov`)

WizTree ([diskanalyzer.com](https://diskanalyzer.com/)) reads the NTFS MFT directly and is roughly the Windows equivalent of `du`. Full write-up in the [Windows comment](https://github.com/unxed/f4/pull/244#issuecomment-5150869085); condensed numbers on `C:\Users\sogonov` (~150 GB, ~550k files, ~106k folders):

**Non-admin (all three utilities under the same rights):**

| | f4 | WizTree | Explorer |
|---|---|---|---|
| Folders | 106 365 | 106 385 | 106 365 |
| Files | 548 631 | 548 615 | 548 616 |
| Files size | 148 840 889 603 | 148 844 785 779 | 148 845 767 001 |
| Physical | 147 546 113 538 | 148 336 754 688 | 148 344 586 240 |

Folders, Files, Files size all match to within noise. Physical is ~790 M lower in f4 — `GetCompressedFileSizeW` returns `INVALID_FILE_SIZE` for system-owned files without `SeBackupPrivilege` and f4 falls back to `info.Size()`; WizTree and Explorer see full allocation for those files.

**Admin — snapshot BEFORE the two admin-Windows fixes below** (only f4 admin vs WizTree admin — Explorer wasn't re-run under admin):

| | f4 (admin, pre-fix) | WizTree (admin) |
|---|---|---|
| Folders | 108 678 | 106 926 |
| Files | 570 639 | 554 501 |
| Files size | 150 564 868 508 | 149 494 904 779 |
| Physical | 149 267 912 451 | 148 994 588 672 |

Two admin-Windows issues showed up on this tree and are both fixed in this PR:

* **Runaway recursion (fixed earlier in this PR).** An earlier build reported 3.12M files / 526 GB physical on a 340 GB partition, because Go's DirEntry.Type() didn't always mark NTFS junctions as `ModeSymlink` and the scanner walked into `C:\Users\<user>\AppData\Local\Application Data → parent` self-loops. Fixed by the `isReparsePoint` + `FILE_ATTRIBUTE_REPARSE_POINT` commit (`4e8ebc6`, "Windows: treat all NTFS reparse points as symlinks in the scanner").
* **~3% residual over-count (also fixed).** After the junction fix, the numbers above still ran ~+16k files / +270 M physical vs WizTree admin. Root cause was hard-link dedup being a no-op on Windows — `VFSItem.Device`/`Inode` weren't populated because `FindNextFile` doesn't return them. Fixed by two follow-up commits:
    * `b0067ae` "Windows: resolve file identity so the scanner dedups hard links" — new optional `FileIdentifier` VFS capability. Scanner resolves identity **lazily** and **only under `DedupInodes`** (so ordinary panel listings pay no extra syscall), OSVFS implements it via `GetFileInformationByHandle` on Windows (volume serial + 64-bit file index) and `Lstat` on Unix.
    * `1d82a64` "Windows: report real on-disk size and drop the root dir from Files size" — two small Windows-specific tweaks: `Physical size` now uses `GetFileInformationByHandleEx(FileStandardInfo).AllocationSize` (cluster-granular, compression/sparse aware — the true Windows analogue of `stat.Blocks*512`); `Files size` drops `DirBytes` on Windows because `os.Stat` reports 4096 for a directory index size which would leak into the total. Verified on a separate 116 GB tree (`D:\sprint2`): Files size and Physical now match Far/WizTree.

WizTree admin also runs noticeably faster than under user rights (single MFT sweep instead of per-file `GetCompressedFileSize`). f4 speed is comparable to Explorer's — nowhere near WizTree admin; parity with MFT-level tooling isn't attempted here.

### What's out of scope, follow-up PRs

* **far2l's `+3 files / +435 M` on `~/` and `+38 files / +150 M` on `/home/linuxbrew`** — reproduced but not attributed. Probably worth its own thread in the far2l repo; deltas match neither `find` / `du` on Linux nor WizTree / Explorer on Windows.
* ***BSD** (dragonfly / netbsd / openbsd / illumos / solaris / freebsd) — not measured. `fsInfo` there is still a stub so Cluster size doesn't render; scanner side works out of the box through the unix build path.
* Live refresh when directory contents change while the panel is open (far2l behaves the same way — you have to move the cursor or reopen Ctrl+Q to re-scan).
* `FastScanner` PhysicalBytes for remote VFSes.

In short — Ctrl+Q on a typical directory now gives numbers you can trust and cross-check against neutral tools on both Linux and Windows; the remaining edge cases are catalogued above and easy to reproduce.


sogonov:
## Reproducible sanity-check of the numbers Ctrl+Q shows

To keep everyone honest about which tool is right on which metric, here's a self-contained shell recipe. Runs three independent implementations of "recursive size" against the same tree and prints their answers side-by-side; anyone can paste this at a Linux prompt.

```bash
# Any directory works; big trees just take longer.
TARGET=/home/sogonov

# 1. find — the reference for entry counting and (with dedup) physical size.
#    -printf '%y %b %i' → type, blocks*512-units, inode.
find "$TARGET" -printf '%y %b %i\n' 2>/dev/null | awk '
BEGIN {fc=dc=lc=sc=pc=bc=cc=0}
$1=="f" {fc++; if(seen[$3]++) dupf++; else phy+=$2}
$1=="d" {dc++; if(seen[$3]++) dupd++; else phy+=$2}
$1=="l" {lc++; if(seen[$3]++) dupl++; else phy+=$2}
$1=="s" {sc++}
$1=="p" {pc++}
$1=="b" {bc++}
$1=="c" {cc++}
END {
  raw = fc + lc + sc + pc + bc + cc
  printf "  raw non-dir entries : %d  (regular=%d links=%d sockets=%d pipes=%d blk=%d char=%d)\n", raw, fc, lc, sc, pc, bc, cc
  printf "  hard-link duplicates: %d  (files with a repeated inode)\n", dupf
  printf "  after dedup         : %d  ← compare to Files in Ctrl+Q\n", raw - dupf
  printf "  dirs (raw)          : %d\n", dc
  printf "  physical            : %d bytes  (blocks*512, dedup by inode)\n", phy*512
}'

# 2. du — the second, independent source of physical truth.
du -sk "$TARGET" 2>/dev/null | awk '{printf "  du -sk              : %d bytes\n", $1*1024}'
```

Sample output on my `/home/sogonov` (a ~15 GB tree, ~125k files):

```
  raw non-dir entries : 126 937  (regular=126691 links=79 sockets=167 pipes=0 blk=0 char=0)
  hard-link duplicates: 917  (files with a repeated inode)
  after dedup         : 126 020  ← compare to Files in Ctrl+Q
  dirs (raw)          : 13 496
  physical            : 15 582 568 448 bytes  (blocks*512, dedup by inode)

  du -sk              : 15 582 568 448 bytes
```

`find %b dedup` and `du -sk` matched **bit-for-bit** — that's the ground truth.

Ctrl+Q on the same tree in f4 (this PR): **Files 126 020, Physical 15 582 261 248**. Files bit-for-bit, Physical within a few hundred KB of `du` between snapshots. A later synchronous run reproduced **f4 = `du` bit-for-bit** on Physical (`16 403 861 504` on both) — no residual drift on our side.

Ctrl+Q on the same tree in far2l: **Files 126 023 (+3 vs find), Physical 16 017 948 672 (+435 MB vs du)**. Both deltas are systematic — reproduced across three separate runs — not noise. This looks like a far2l issue worth filing separately; ticket-worthy material is already assembled in the PR description.

## Other "sources of truth" for cross-checking on Linux / *BSD

Every one of these should agree with `du` on physical (they all sum stat.blocks, dedup by inode). Handy if `du` or `find` isn't available or if you want a visual answer:

* **`ncdu`** — the classic curses TUI. `ncdu /home/sogonov` — shows per-directory sizes; totals in the header match `du`.
* **`gdu`** — Go port, much faster on big trees; same numbers.
* **`dust`** — Rust, prettier UX; same numbers.
* **`diskus`** — Rust, minimalist single-line answer.
* **`tar --list --totals`** — bytes only, no physical; useful for logical size sanity.
* **`stat -c %b` in a loop** — for one-off manual verification.
* On btrfs / zfs specifically, `btrfs filesystem du` / `zfs list -o used` reflect the compressed on-disk footprint if you want to double-check transparent-compression cases.

The intent isn't to add any of these to CI — just so a reviewer with a different tool at hand can confirm f4's numbers independently.

## What's in the PR now vs what stayed on the far2l side

* Bit-for-bit match with `find` on Files/Dirs and with `du` on Physical: **done**.
* Layout / labels / units toggle / hard-link dedup / symlink-leaf / permission-denied recovery: **all in this PR**.
* The extra 3 files and 435 MB physical that far2l shows on this tree: **outside f4** — nothing in the scanner or the panel here inflates by those amounts; three independent tools (`find`, `du`, this PR) agree.

sogonov:
## Windows sanity-check против WizTree (независимая правда для NTFS)

Ссылка на утилиту: <https://diskanalyzer.com/> (Antibody Software, WizTree — читает NTFS MFT напрямую и часто используется как «золотой стандарт» для recursive size на Windows).

Замер на `C:\Users\sogonov` (~150 GB, ~550k файлов, ~106k папок).

### Non-admin (все три утилиты под одинаковыми правами)

| | f4 | WizTree | Explorer |
| --- | --- | --- | --- |
| Folders | 106 365 | 106 385 | 106 365 |
| Files | 548 631 | 548 615 | 548 616 |
| Files size | 148 840 889 603 | 148 844 785 779 | 148 845 767 001 |
| Physical | 147 546 113 538 | 148 336 754 688 | 148 344 586 240 |

Folders, Files, Files size — в пределах шума / нескольких MB. **Physical у f4 ниже на ~790 M**: WinAPI под user rights возвращает ошибку для system-owned файлов без `SeBackupPrivilege`, мы честно откатываемся на `info.Size()`. WizTree и Explorer видят полный allocation для тех же файлов.

WizTree в этом режиме показывает баннер «нужны admin-права для полного сканирования».

### Admin

Замер **ПОСЛЕ** двух Windows-admin фиксов из этого PR (см. ниже):

| | f4 (admin) | WizTree (admin) | Far (admin, D:\sprint2) |
| --- | --- | --- | --- |
| Files size (D:\sprint2, verified sample) | 116 830 248 699 | matches | matches |
| Physical (D:\sprint2, verified sample) | 116 854 410 896 | matches | ~0.2 MB above (Far округляет до кластера) |

**До** фиксов на `C:\Users\sogonov` было два admin-only issue:

1. **Runaway recursion**: f4 admin показывал 3.12M файлов / 526 GB на 340 GB partition — сканер зацикливался через junctions вроде `AppData\Local\Application Data → parent`. Go иногда помечал их не как `ModeSymlink`. Fix — commit `4e8ebc6` "Windows: treat all NTFS reparse points as symlinks in the scanner" (использует `FILE_ATTRIBUTE_REPARSE_POINT` напрямую).

2. **~3% residual over-count** относительно WizTree admin (+16k файлов, +270 M physical). Причина — на Windows не работал hard-link dedup: `VFSItem.Device`/`Inode` пусты, потому что `FindNextFile` их не отдаёт. Fix:
    * `b0067ae` "Windows: resolve file identity so the scanner dedups hard links" — опциональный `FileIdentifier` VFS capability, scanner дёргает `GetFileInformationByHandle` **лениво и только под `DedupInodes`** (обычный листинг панели не платит).
    * `1d82a64` "Windows: report real on-disk size and drop the root dir from Files size" — Physical через `GetFileInformationByHandleEx(FileStandardInfo).AllocationSize` (cluster-granular, compression/sparse aware, аналог `stat.Blocks*512` на Unix) + Files size = `stats.Bytes` без DirBytes на Windows (`os.Stat` для директории возвращает 4096 index size, что течёт в total).

WizTree admin также **заметно быстрее** — один проход MFT вместо per-file `GetFileInformationByHandle`. По скорости f4 сопоставим с Explorer; сравняться со WizTree под admin — это отдельный подход (MFT-parser), не в scope этого PR.

## Как читать имеющиеся замеры

* **Non-admin — f4 = WizTree non-admin = Explorer** по счётчикам и logical size; physical ниже из-за system-owned файлов, куда стандартный WinAPI без admin не пускает. Ожидаемое поведение.
* **Admin — после фиксов** f4 admin даёт числа, совпадающие с WizTree/Explorer на sample tree; runaway recursion и residual over-count оба закрыты.

## Что дальше — отдельным PR

Основное для типового Ctrl+Q под user rights уже работает; открытые вопросы:

* **f4 non-admin: Physical ~790 M ниже WizTree/Explorer** из-за `INVALID_FILE_SIZE` fallback на protected системные файлы. Fix потребует либо auto-elevation через `SeBackupPrivilege`, либо MFT-парсера (значительный кусок работы).
* **macOS и *BSD** — macOS уже cross-verified (`f4 = du` bit-for-bit); *BSD не измерены, `fsInfo` там всё ещё стуб.

sogonov:
## Windows: доводка Physical / Files size + hard-link dedup (в этом PR)

Windows follow-up'ы из [комментария выше](https://github.com/unxed/f4/pull/244#issuecomment-5150869085) закрыты прямо в этом PR — тремя нативными правками. Проверял на чистом стабильном дереве `D:\sprint2` (167 папок, 11 864 файла) против Проводника и WizTree.

### Что изменилось

1. **Physical size → реальная аллокация.** Было `GetCompressedFileSizeW` — она возвращает «упакованный» размер данных **без округления до кластера**, поэтому занижала footprint на кластерные остатки (~24 МБ на 116 ГБ) и расходилась с Unix-веткой (`stat.Blocks*512`). Стало `GetFileInformationByHandleEx(FileStandardInfo).AllocationSize` — cluster-granular, с учётом NTFS-compression / sparse / MFT-resident. Каталоги дают 0 (как Far «Реальный размер»).
2. **NTFS hard-link dedup на Windows.** Добавлена optional VFS-capability `FileIdentifier`; сканер лениво разрешает `(Device, Inode)` через `GetFileInformationByHandle`, **только под `DedupInodes`** (guard `seen != nil`) — обычный ReadDir/листинг per-file-syscall не платит, той регрессии, что вычищали из ReadDir path, не возвращается. Это и есть закрытие «admin over-count» из follow-up.
3. **Files size больше не учитывает сам сканируемый каталог.** `os.Stat` папки на Windows возвращает 4096 (размер её индекса), и это единственный ненулевой вклад в `DirBytes` (дочерние каталоги через ReadDir = 0). На Windows логический размер теперь = только файловые байты (как Проводник/Far); на Unix оставлено `Bytes + DirBytes` (паритет far2l).

### Сверка на `D:\sprint2` (admin)

| | f4 (было) | **f4 (стало)** | Проводник | WizTree | Far 2.0 |
| --- | --- | --- | --- | --- | --- |
| Folders | 167 | 167 | 167 | 167 | 167 |
| Files | 11 864 | 11 864 | 11 864 | 11 864 | 11 864 |
| Files size | 116 830 252 795 | **116 830 248 699** | 116 830 248 699 | 116 830 248 699 | 116 830 248 699 |
| Physical | 116 830 248 699 | **116 854 410 896** | 116 854 398 976 | 116 854 390 784 | 116 854 636 544 |

* **Files size — точное совпадение** с Проводником / WizTree / Far (ушли лишние +4096 корня).
* **Physical** теперь в «семье реальной аллокации»: f4 / Проводник / WizTree сходятся в пределах **~20 КБ** (0.00002 % от 116 ГБ), а наивный Far (`Σ ceil(size/cluster)*cluster`) выпадает на +226 КБ, т.к. игнорирует сжатие. Побайтно три «allocation»-инструмента не совпадают никогда — расходятся на пограничных случаях (индексы каталогов, гранулярность аллокации сжатых файлов, reparse/ADS, hard-link dedup).

### Правки
- `vfs/os_vfs_physical_windows.go`, `vfs/os_vfs_physical_unix.go`, `vfs/os_vfs_physical_other.go`, `vfs/scanner.go`, `quick_view_panel.go`.
- `go vet ./vfs/` чист; целевой тест `TestGenericScan_HardLinkDedup` проходит на Windows.

Полноценный MFT-parser ради скорости WizTree под admin по-прежнему вне scope этого PR.

PR #287, opened 2026-08-03
Add cross-platform app icons and Windows integration

Zoinen:
Generate size-specific platform assets, package them in CI, and apply DPI-aware Windows icons and title-bar themes.


unxed:
Ух ты, здорово как!

PR #288, opened 2026-08-03
Add search-first panel navigation mode

Zoinen:
## Summary
- replace the Vim hotkey toggle with Classic, Vim, and Search First navigation modes
- route printable input to Fast Find in Search First and toggle command-line focus with the grave key (Quake-like)
- add inactive panel cursor and command prompt colors plus settings and config migration
- map Detailed view left/right navigation to Page Up/Page Down outside Vim mode
- allow Alt+grave to insert a literal backtick when the command line is focused

## Tests
- go test . -run 'Test(SearchFirst|Detailed|Navigation|Config|Settings)|TestPanelNavigation' -count=1


unxed:
Поправь конфликты плз

PR #292, opened 2026-08-03
Add four panel view modes and extension alignment

Zoinen:
## Summary
- add Brief, Medium, Detailed, and Wide panel modes with Ctrl+1…4 shortcuts (just like in Far)
- preserve normal per-panel modes while Wide occupies the full panel width
- persist Wide state while keeping existing session.ini mode values compatible
- add a panel setting that aligns extensions to the right inside the Name cell
- update English/Russian localization, semantic names, and regression tests

## Testing
- targeted panel mode, layout, session, configuration, and extension formatting tests
- go build -o f4.exe .


unxed:
по RCtrl+цифрам работали закладки, а теперь тоже смена режима

поправьте плз

PR #295, opened 2026-08-04
Panels: Shift-nav range selection, ESC toggle panels (issue #289)

sogonov:
Addresses the Panels bullets from #289:

* **перемещение с нажатым Shift должно выделять все файлы** — Shift+navigation now paints the whole swept range, not just the starting row. Works uniformly for Up/Down/Left/Right/PgUp/PgDn/Home/End.
* **ESC — включение/выключение панелей** — ESC toggles panel visibility (Ctrl+O analog).

## What each commit does

1. **`FilePanel: Shift+navigation with additive multi-step selection`** — first cut: Shift+multi-step (Left/Right/PgUp/PgDn/Home/End) paints the whole swept range as \"selected\"; Shift+Up/Down keep FAR's per-tap toggle-current-only. Also drops the ".." parent-dir special-case worry (SetItemSelected already skips it).
2. **`FilePanel: drop persistent selection when the directory changes`** — pre-existing bug on main: \`selectedItems\` (map keyed by bare filename) survived across directory changes, so a file selected in ~/f4 (say \`.claude\`) came back pre-selected in ~/scc and ~. Fixed by tagging the map with \`lastLoadedPath\` in \`readDirectoryEx\`; \`Clone()\` copies the tag alongside \`selectedItems\` so cloning a workspace doesn't trigger the drop.
3. **`FilePanel: Shift+nav session mode — unified select/deselect`** — user feedback loop: after a Shift+swipe the mode was always \"add\"; couldn't shift-deselect. Introduces session semantics: while Shift is held, the mode (select vs deselect) is decided on the first Shift+nav from the state of the row under the cursor, and every subsequent Shift+nav in the same session applies that same mode. Releasing Shift or pressing any non-nav key closes the session; the next Shift+nav re-decides. Uniform across all nav keys. \".\" starting row falls back to the first real row in the direction of movement, so Shift+End from \".\" over a fully selected list correctly picks deselect.
4. **`PanelsFrame: ESC toggles panels visibility`** — mirrors FAR's shipped-with-standard-install ESC macro. Guarded so ESC still goes to a running interactive app (vim, less, htop) when panels are hidden and the terminal is busy, and still clears the command line when it has a typed command.
5. **`PanelsFrame: gate ESC-toggle behind AppConfig.EscTogglePanels`** — same shortcut is a macro in FAR, so it's opt-outable there. Same choice here: new \`AppConfig.EscTogglePanels bool\`, default \`true\` (matches FAR's stock macro), persisted to \`[Panel] EscTogglePanels\`. New checkbox in F9 → Panel settings.

## Test plan

- [x] \`go test ./...\` passes (\`TestPlugRing_InstallAndRemove_EndToEnd\` fails identically on \`main\` — pre-existing, not this PR).
- [x] \`go build ./...\` clean on linux; \`GOOS=windows GOARCH=amd64 go build ./...\` clean; \`GOOS=darwin GOARCH=arm64 go build ./...\` clean.
- [x] \`gofmt -w -s\` clean on touched files.

### Automated (new)

* \`TestFileSystemPanel_MultiSelect\` — single-step Shift+Down toggles only the starting row.
* \`TestFileSystemPanel_ShiftMultiStepSwipe\` — start on \".\" still paints the sweep; return sweep over already-selected rows doesn't deselect (additive within a session).
* \`TestFileSystemPanel_ShiftRangeSkipsParentDir\` — \".\" is never in the selection.
* \`TestFileSystemPanel_ShiftSessionModeDecidedOnFirstKey\` — full flow: session-1 selects across two taps; plain Down closes it; session-2 starting on a selected row switches into deselect mode.
* \`TestFileSystemPanel_ShiftSessionDeselectFromParentDir\` — \".\" start looks past to the next real row so Shift+End can clear a pre-selected list.
* \`TestFileSystemPanel_SelectionClearedOnDirChange\` — sibling directories with same filename don't inherit selection.
* \`TestPanelsFrame_Clone_SelectionPreservation\` — regression guard (existed on main): the dir-clear commit doesn't strip selection out of a fresh workspace clone.
* \`TestPanelsFrame_EscTogglesPanels\` — hide/show/hide-with-typed-command flow.
* \`TestPanelsFrame_EscTogglePanels_RespectsOption\` — the flag actually turns the shortcut off.

### Manual

Tested end-to-end in WSL Ubuntu (\`f4-linux\`) and native Windows 10 (\`f4.exe\`), including the whole Shift+nav interaction loop across Up/Down/PgUp/PgDn/Home/End/Left/Right (grid), directory switches, workspace clone, ESC-hide with and without a typed command, ESC-restore, and ESC-passthrough into a busy PTY (\`sleep 30\` inside f4's terminal keeps ESC away from the panels toggle).

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -1700,7 +1700,7 @@ func actionFindFile(pf *PanelsFrame) {
 	vtui.FrameManager.Push(dlg)
 }
 func actionPanelSettings(pf *PanelsFrame) {
-	dlg := vtui.NewCenteredDialog(60, 27, Msg("PanelSettings.Title"))
+	dlg := vtui.NewCenteredDialog(60, 28, Msg("PanelSettings.Title"))
 	dlg.ShowClose = true
 
 	chkHidden := vtui.NewCheckbox(0, 0, Msg("PanelSettings.ShowHidden"), false)
@@ -1755,6 +1755,12 @@ func actionPanelSettings(pf *PanelsFrame) {
 		chkCPUGPU.State = 1
 	}
 
+	chkEscToggle := vtui.NewCheckbox(0, 0, Msg("PanelSettings.EscTogglePanels"), false)
+	chkEscToggle.State = 0
+	if AppConfig.EscTogglePanels {
+		chkEscToggle.State = 1
+	}
+
 	modes := []string{Msg("Op.Queue"), Msg("Op.Background"), Msg("Op.Foreground")}
 	comboMode := vtui.NewComboBox(0, 0, 30, modes)
 	comboMode.DropdownOnly = true
@@ -1782,6 +1788,7 @@ func actionPanelSettings(pf *PanelsFrame) {
 	dlg.AddItem(chkSync)
 	dlg.AddItem(chkAlwaysMenu)
 	dlg.AddItem(chkCPUGPU)
+	dlg.AddItem(chkEscToggle)
 	dlg.AddItem(lblMode)
 	dlg.AddItem(comboMode)
 	dlg.AddItem(lblPath)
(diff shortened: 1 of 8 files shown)


sogonov:
Force-pushed one additional commit (`65241df`) that fixes the CI regression: `TestPanelsFrame_HistoryNavigation_HiddenPanels` failed on linux/windows amd64/arm because the ESC-toggle branch was consuming ESC even when the command line had a typed command (typed here via history navigation Up), so the pre-existing "ESC clears cmdLine + resets HistoryPos" handler never got the event.

Fix hoists the `pf.cmdLine.IsEmpty()` check into the outer guard so both branches (hide-visible and show-hidden) require an empty command line. On a non-empty line ESC falls through unchanged. Kept as its own commit rather than squashed into the original ESC-toggle commit so the delta is auditable.

All three EscToggle-related tests plus the regressed one now pass:

```
--- PASS: TestPanelsFrame_EscTogglesPanels (0.00s)
--- PASS: TestPanelsFrame_EscTogglePanels_RespectsOption (0.00s)
--- PASS: TestPanelsFrame_HistoryNavigation_HiddenPanels (0.00s)
```

PR #351, opened 2026-08-06
Improve panel dialogs and path interactions

Zoinen:
## Summary
- close panel fast-find when navigating folder history
- resolve relative Copy/Move destinations from the active source panel
- focus Delete by default while preserving the safety setting
- apply dialog theme colors to Operations Queue
- open a panel drive menu from its path context click or Ctrl+Shift+Left/Right
- insert left/right panel paths into focused dialog edits with Ctrl+[/Ctrl+]
- open path and sort context menus for the clicked panel without activating it
- ignore local profile, settings, and GUI build artifacts

## Tests
- targeted delete, file operation, folder history, panel context menu, drive menu, edit hotkey, and queue theme tests
- `go build -o f4.exe .`


unxed:
Падает сборка на Винде

PR #360, opened 2026-08-06
File associations dialog (fixes #345)

### LINKED ISSUE #345: [feature] far-like custom associations for execute, view and edit
Проще картинкой из фар

Image

Возможность указывать для конкретных масок файлов альтернативные варианты открытия, просмотра и редактирования (или оставлять дефолтные)

sogonov:
Fixes #345.

Ports far2l's per-mask file associations feature. Files matching a user-defined mask can override Enter (open), F3 (view) and F4 (edit) with an arbitrary shell command. When two or more associations match the same action, a picker menu appears — otherwise the command runs directly. When nothing matches, the built-in handler runs as before.

## Data model & storage

- `FileAssoc` mirrors far2l's `FileTypeStrings`: six command slots per the `FILETYPE_*` enum (`Execute` / `AltExec` / `View` / `AltView` / `Edit` / `AltEdit`). Enable bits carry the far2l `State` bitmask. Only Execute / View / Edit are wired to hotkeys today; the four `Alt-` slots are preserved on disk so an `associations.ini` authored in far2l round-trips byte-for-byte.
- Storage: `~/.config/f4/settings/associations.ini`, sections `[Associations/Type%d]`, stop-at-first-gap loader (matches far2l). Atomic tmp+rename saves.

## Mask matching

`MatchMask` supports the full far2l `CFileMask` syntax:

- `,` and `;` — OR-separators: `*.png,*.jpg,*.gif`
- `|` — exclude side: `*.png,*.jpg|thumb_*` (png or jpg, but not thumb_*)
- `/…/` — a regex segment: `/^Makefile.*$/`
- Case-insensitive matching honoured for both globs and regex (via `(?i)`)

Malformed masks never match — safer than degrading to "matches everything".

## Dispatch

`tryFileAssociation()` is called at the top of `actionExecute`, `actionViewFile` and `actionEditFile`. A single match runs directly through `executeMenuCommands`, so token substitution (`!.!`, `!\!`, `!?…!`, everything `user_menu_subst` already handles), history, PTY wiring, OSC 133 and terminal muting come for free. Multiple matches push a `VMenu` picker; Enter on a row dispatches the picked command. No match → the intercept returns false and the existing default fires unchanged.

## Editor UI

F9 → Commands → **File Associations…** opens a list dialog:

- **Ins** — add
- **Del** — remove (with confirmation)
- **F4** / **Enter** — edit selected
- **Ctrl+Up** / **Ctrl+Down** — reorder in place

Per-entry editor has Mask, Description and one (checkbox, edit) row per slot. Same chord vocabulary as the existing user-menu editor.

## Tests

- Mask matcher: basic globs, ignore-case, exclude, regex (including malformed).
- INI round-trip, missing file, stop-at-first-gap, State bitmask expansion.
- Dispatch: single match runs, no match falls through, disabled slot skipped, wrong-kind slot skipped, directories never intercepted, multi-match shows picker, picker Enter dispatches the picked command, exclude vetoes match.
- Editor: fill-fields-Save round-trip through the real dialog.
- Lang keys: every string added to `lang/en.lng` and `lang/ru.lng` resolves.

Full `go test ./...` is green locally.


unxed:
Спасибо! Пара тестов только падает

sogonov:
Ребейзнул на актуальный main (bea2f5b) и разобрался с падавшим тестом.

**TestGenerateKeysHelpTopic_PanelNav** — это унаследованный с main баг: PR #351 добавил `Ctrl+Shift+Left/Right` алиасами к `Alt+F1/F2` (Zoinen's panel-dialog-interactions), но не обновил тест — тот всё ещё ждал старую строку `"Alt+F1         - Show the drive menu for the left panel"`, а `generateKeysHelpTopic` теперь генерирует `"Alt+F1 / Ctrl+Shift+Left - Show the drive menu for the left panel"`. Проверил чистый `upstream/main` в отдельном worktree — тест падает без моих изменений. Заодно поправил ожидаемую строку в тесте — комментарий в коде объясняет, что и почему.

`go test ./...` локально зелёный. Пуш force-with-lease, HEAD теперь `c685399`.

PR #365, opened 2026-08-07
perf(image): use woozymasta/png for PNG decode; upgrade klauspost/compress

Procrustean:
faster-png: route PNG decode through an optimized zlib-based decoder, faster decode with byte-identical output, no API change.
Decode speedup vs stdlib on real corpora (photo 27 MB, binary, source text, UI screenshot): ~22–37% faster, output pixels verified byte-identical.


unxed:
Cool, thanks!

PR #418, opened 2026-08-09
plugins/netfox/fishplus: PowerShell helper for Windows peers

sogonov:
Windows counterpart of helper.sh: a plain PowerShell 5.1+ script that speaks the FISH+ wire protocol v1 against a Windows OpenSSH-Server peer. Same client, same VFS, same test lab pattern — POSIX peers keep using helper.sh, the client picks between them based on how the transport handshake answers.

Transport in netfox/fish_vfs.go tries POSIX first ('exec /bin/sh' through the login shell), and on any handshake failure — an EOF from a cmd DefaultShell that could not parse the request, a timeout from a peer that stayed silent, an 'unexpected banner' from a shell that answered but does not speak FISH+ — falls back to a PowerShell shell request. The fallback runs 'powershell.exe -NoLogo -NoProfile' via sess.Start, so a stock OpenSSH-Server-Windows with DefaultShell still at cmd.exe routes correctly through cmd /c powershell. On a Windows peer whose DefaultShell is already powershell the request nests, at the cost of a second launch but with the same wire result.

Notable choices reflected in helper.ps1 (details in WINDOWS_PORT.md):

- stdin comes through $host.UI.ReadLine() re-encoded with the console code page, not through [Console]::OpenStandardInput() — the console host holds a pending read on the raw handle and swallows bytes the script would otherwise see.

- Content search (grep, ffind) reads 1 MB chunks decoded through ISO-8859-1 (byte-transparent) and matches with String.IndexOf / Regex.IsMatch on that string. Behaves like 'grep -a': RSS stays at the buffer size regardless of file size or binariness, and a Cyrillic pattern still matches its own UTF-8 bytes.

- exec jobs bypass Start-Job and run cmd.exe directly through a small wrapper batch that redirects stdin from NUL and stdout+stderr into the job's output file. Start-Job under the cmd-launched PS 5.1 hangs the whole session in ways that are not worth debugging.

- scan and hash jobs use [PowerShell]::Create() + BeginInvoke() — in-process runspace, same priority as the main helper — instead of Start-Job's separate pwsh.exe process (which gets background I/O priority and slows content-shaped work by up to 100×).

- Reparse points are never followed during recursion: Windows profile trees have self-referential junctions ('AppData\\Local\\ Application Data' -> 'AppData\\Local') that make every recursive .NET or PowerShell walk loop forever otherwise.

Feature announcement extends the base FISHPLUS 1 banner with 'flavor:pwsh', 'mode:stat' (helper.ps1 emits stat-format entries so the client's existing parseListing path is used unchanged), 'hash:sha256sum' (gates the duplicate search), 'read:filestream' and 'write:b64' — with everything the client's CanX gates need to open the corresponding features.

Tests:

- session_pwsh_test.go integration-runs the whole helper against a local pwsh, gated on pwsh presence + non-Windows (Windows console pipes deadlock the write-then-read handshake in a subprocess).
- script_pwsh_test.go checks the Base64BootstrapLinePwsh producer is single-line, printable ASCII, and decodes back to the compact helper with the session token substituted.

ffind stays synchronous in this PR (the sync path helper.sh has). A following PR replaces it with a background job kind and adds a parallel worker pool, keeping the transport contract the same.

### DIFF
--- a/plugins/netfox/fish_vfs.go
+++ b/plugins/netfox/fish_vfs.go
@@ -72,8 +72,17 @@ type FishDialer func(ctx context.Context) (io.Writer, io.Reader, io.Closer, erro
 type fishConn struct {
 	client *fishplus.Client
 	ka     *fishplus.Keepalive
-	dial   FishDialer
-	closer io.Closer
+	// dial and opts describe how to bring the primary shell flavor up. A
+	// site the first attempt cannot reach — a Windows peer whose sshd
+	// answers "exec /bin/sh" with a PowerShell parse error — is served by
+	// dialAlt/optsAlt instead. Once a fallback succeeds it becomes the
+	// primary, so every reconnect after the first goes straight to the
+	// flavor that actually works.
+	dial    FishDialer
+	opts    fishplus.HandshakeOptions
+	dialAlt FishDialer
+	optsAlt fishplus.HandshakeOptions
+	closer  io.Closer
 
 	mu     sync.Mutex
 	refs   int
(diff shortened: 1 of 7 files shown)


unxed:
Фантастика!

В FISH+.md надо пункт Support for Windows hosts перенести в готовое, наверное?
UPD: а, лан, я сам

PR #436, opened 2026-08-10
Streamline the default panel layout

Zoinen:
## Summary
- add a panel setting for the separate current-file information row
- hide that row and its separator by default, reclaiming two rows for file entries
- keep selection and directory totals visible on the bottom panel border regardless of the setting
- make the workspace tab bar always visible by default while preserving saved user preferences
- add English and Russian labels plus persistence and rendering regression tests

## Tests
- targeted config, panel rendering, layout, and settings-dialog tests pass
- `go build -o f4.exe .` passes

## Known unrelated Windows test issues
The full main-package run currently encounters temporary-file locks in the existing Viewer/Editor tests and the existing `TestApplyCompletionDoesNotTouchClosedWorkspace` race.


unxed:
Чот там падает ещё какая-то сборка

PR #443, opened 2026-08-11
editor: add Find All and overhaul search/replace matching

enedzvetsky:
## Find All (Far's find_all_list)

The search dialog's **[ All ]** button collects every occurrence into a menu titled "Occurrences: N, lines: M" with line/col prefixes and the match highlighted inside each item.

- **Enter** jumps to the occurrence, **Ctrl+Enter** jumps while keeping the menu open
- **Ctrl+Up/Down** scroll the editor behind the menu
- **F4** dumps the matching lines into a new editor, **F5** zooms the menu
- Collection runs in the background with a cancelable progress dialog; results are dropped if the buffer was edited meanwhile
- Handles terminal resize while open (including the zoomed state)

## Search/replace matching rebuilt around charlievieth/strcase

Case-insensitive literal search used to run over a `strings.ToLower` copy of the buffer. Offsets found in that copy drift as soon as any rune changes byte length when lowered (`İ` U+0130, Kelvin sign U+212A): search jumped to the wrong place, and replace spliced the buffer at the wrong offsets, corrupting text around such characters.

[strcase](https://github.com/charlievieth/strcase) folds while scanning the original text, so offsets are exact by construction, with `strings.EqualFold` semantics, zero allocations, and no full-buffer copies. The regex engine is now used only for actual regex and whole-word modes (benchmarked: a `(?i)` literal scan through the regex engine is ~15-20x slower than strcase on match-free buffers).

## Replace fixes

- **Replace with no active selection only selected the first match** - the button appeared to do nothing, since replacing required reopening the dialog with the selection intact. One click now finds, replaces, and highlights the next occurrence (Far-style, one occurrence per invocation).
- The follow-up search no longer skips a match starting immediately after the replacement (e.g. `aa` -> `x` in `aaaa`).
- Replacement text is literal in non-regex modes: `$1` expands only when the pattern was typed as a regex.
- Render state is invalidated from the edited line rather than the cursor line, fixing stale rendering for reverse-direction replaces above the cursor.
- The "Replace with" field remembers its last value in the session, like the search pattern.

## Tests

Unit tests for span collection (plain/regex/whole-word/case-folded, Unicode offset regressions, folded match lengths, cancellation) and menu behavior (jump, Ctrl+Enter, scroll, F4, F5 zoom, resize), plus a regression test for the one-click Replace flow.


unxed:
Great job, thanks!

PR #445, opened 2026-08-11
editor: Find All and interactive search/replace overhaul

enedzvetsky:
## Summary

Two commits building out the editor's search/replace machinery:

**Find All + matching overhaul** (6bfec25)
- New Find All results menu (zoomable, resize-aware, per-line grouping) built on a shared `findAllMatchSpans` scan
- Search/replace matching unified on `buildSearchRegex`/`findMatch`: whole-word via `\b` wrapping, case folding via `strcase` (offset-safe for multi-byte folds like U+212A), reverse regex search without byte-offset corruption

**Interactive Replace confirm loop** (fd160cb)
- Far-style Replace / All / Skip / Cancel dialog per occurrence: the match is scrolled to the top quarter, highlighted, and shown with its concretely rendered replacement ($group expansion only for typed regexes)
- [All] collects the remaining occurrences in one linear `findAllMatchSpans` pass and applies them as per-span splices under a single undo step
- Replace splices register with the editSession/indexer fences (`noteBufferEdit`); background scans are fenced by a session captured while offsets are valid
- Zero-length regex matches are stepped over instead of terminating the loop; skipping every occurrence reports "0 occurrence(s) replaced" instead of "Pattern not found."
- A selection that is exactly one occurrence is prompted as-is instead of re-searched (a regex like `a+` can no longer silently widen the replaced range)
- Shared helpers for replacement rendering (`renderReplacement`) and search seeding (`searchSeedOffset`); new `Search.Title`/`Search.NotFound` lang keys (en/ru)

## Testing

- `go test -run 'TestEditor|TestReplace|TestFindAll|Search' -count=1 .` — all pass
- Interactive flow covered by new tests: prompt flow (Replace/Skip/Cancel), [All] single-undo, adjacent occurrences, replacement-contains-pattern, per-occurrence regex expansion, folded different-length matches, reverse right-to-left walk

### DIFF
--- a/editor_find_all.go
+++ b/editor_find_all.go
@@ -258,7 +258,7 @@ func (ev *EditorView) FindAll(pattern string, caseSensitive, useRegex, wholeWord
 					return // buffer changed while scanning; offsets are stale
 				}
 				if len(spans) == 0 {
-					vtui.ShowMessage(" Search ", "Pattern not found.", []string{"&Ok"})
+					vtui.ShowMessage(Msg("Search.Title"), Msg("Search.NotFound"), []string{Msg("vtui.Ok")})
 					return
 				}
 				ev.showFindAllMenu(pattern, bytes, spans)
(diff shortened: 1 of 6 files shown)


unxed:
Great job, thanks!

PR #455, opened 2026-08-12
editor: Find All, search/replace overhaul, interactive Replace confirm

enedzvetsky:
@
## Summary

- **Find All** in the editor: collects all matches into a navigable results list.
- **Search/replace matching overhaul**: reworked match logic shared by Find, Find All, and Replace.
- **Interactive Replace confirm loop**: per-match Replace / All / Skip / Cancel confirmation, with review fixes.
- **vtui updated to v0.1.183**, which includes the macOS Cmd key fix (Cmd and physical Ctrl are now delivered on separate Ctrl channels).

@


enedzvetsky:
Closing: all changes here (Find All / search-replace work and the vtui v0.1.183 bump) are already in main via #443, #445, and an upstream vtui update.

PR #477, opened 2026-08-15
editor: scan the buffer in place when searching, and fix the in-place save

enedzvetsky:
Two editor fixes, one commit each.

### Search assembles the buffer once, or not at all

Searching a freshly opened large file reported **"Failed to read file buffer"**. `pt.Bytes()` reads every piece in one go and fails as soon as one is still unfetched, which is the normal state right after open — `showEditor` prewarms a single chunk and the rest arrives lazily. `readSearchSnapshot` reads forward in bounded steps and retries on `ErrLoading`, so the reads themselves pull the file in, the way the background indexer does.

Assembling that buffer was then the dominant cost of every search. Each pass copied the file four times over — out of the chunk cache, into `GetRange`'s result, into the assembled snapshot, and once more into a string for the literal matcher — and did it again for the next search over text that had not changed. An interactive Replace paid the whole thing *per confirmation*.

`PieceTable.View` hands out a window into the buffer when the requested range lies inside a single piece, which for an unedited file is the whole of it, so nothing is copied. Buffers that cannot be viewed keep one assembled copy on the editor, keyed by `editSession` so any edit retires it.

Measured on a 200 MB file of 21M lines:

| | before | after |
|---|---|---|
| repeated search | 106 ms, 800 MB allocated | **3.5 ms, none** |
| ten Replace confirmations | 238 ms | **~1 µs** |
| first search on a lazy buffer | 106 ms, 800 MB | 135 ms, 600 MB |

The first search still assembles one copy; a memory-mapped backing buffer would remove that too, and then the cache falls away with it.

### Saving could rename a stage that was never created

Saving a local file could fail with:

```
Failed to finalize save (rename failed):
rename /Users/.../.f4tmp-c820e73d... /Users/.../aa.txt: no such file or directory
```

`OSVFS` implements `InPlacePatcher`, and `SaveToFile` tries it first: it writes the changed pieces straight into the destination and sets `saved`. But `useTemp` stayed set, so the finalize step went on to rename the staged sibling into place — and no sibling had been created, because the write never went through one. The content was already correct on disk; the editor reported a failed save over it and stayed modified, so the next F2 repeated the whole thing.

`PatchInPlace` accepts exactly the edits that leave every unchanged piece at its original offset, so this hit same-length replacements and saving an unmodified buffer, while insertions and deletions shifted the offsets, were refused, and quietly took the staged path that works. That is why it looked intermittent.

### Tests

- `editor_search_zerocopy_test.go` — `View` aliases the buffer rather than copying it and declines across a piece boundary; a search pass over a memory-backed table allocates nothing; a second pass reuses the assembled buffer and an edit retires it.
- `editor_search_lazy_test.go` — the snapshot waits for lazily loaded chunks, stops promptly on cancel, and handles the fully resident and empty cases.
- `editor_save_inplace_test.go` — the in-place save completes, an untouched buffer saves, and an insertion still goes through the staged path. These assert the editor goes clean, because the content check alone cannot see the bug: `PatchInPlace` writes it either way.

Note on the local suite: `TestRecursiveCopy_SymlinkLoop` hangs and three sibling `file_ops` tests fail on macOS — verified identical on a clean `main` worktree, unrelated to this branch.

### DIFF
--- a/editor_find_all.go
+++ b/editor_find_all.go
@@ -231,8 +231,11 @@ func (ev *EditorView) FindAll(pattern string, caseSensitive, useRegex, wholeWord
 		session := ev.editSession
 
 		runSearchWithProgress(pattern, func(ctx *vtui.TaskContext, dlg *vtui.Window) {
-			bytes, errBytes := ev.pt.Bytes()
+			bytes, errBytes := ev.searchBuffer(ctx, session)
 			if errBytes != nil {
+				if ctx.Err() != nil {
+					return // canceled; the dialog is already closing
+				}
 				ctx.RunOnUI(func() {
 					dlg.Close()
 					vtui.ShowMessage(" Error ", "Failed to read file buffer.", []string{"&Ok"})
--- a/editor_replace_confirm.go
+++ b/editor_replace_confirm.go
@@ -90,9 +90,13 @@ func selectionIsMatch(sel []byte, pattern string, caseSensitive bool, re *corege
 func (st *replaceLoop) findNext() {
 	ev := st.ev
 	vtui.FrameManager.PostTask(func() {
+		session := st.session
 		runSearchWithProgress(st.pattern, func(ctx *vtui.TaskContext, dlg *vtui.Window) {
-			data, errBytes := ev.pt.Bytes()
+			data, errBytes := ev.searchBuffer(ctx, session)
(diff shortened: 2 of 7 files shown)


unxed:
Great job, thanks!

PR #478, opened 2026-08-15
pty: mark master and slave close-on-exec so shells die with f4

enedzvetsky:
## The problem

A PTY allocated by f4 keeps its shell alive **forever** after f4 itself is gone, and each leaked shell holds a pty node. Enough restarts walk the host into its pty limit, after which allocation fails with `ENXIO` and no terminal can be opened anywhere on the machine — not f4's, not the system terminal's.

This is not theoretical. It was found on a host that had accumulated **497 orphaned shells against a `kern.tty.ptmx_max` of 511**, all spawned in one burst and idle for two days, each pinning a `/dev/ttys` node:

```
$ ps -eo tty,comm | awk '$1 ~ /^ttys/ {print $2}' | sort | uniq -c | sort -rn
 499 /bin/zsh
   5 claude
```

## Root cause

The descriptors are opened with a raw `unix.Open` and wrapped in `os.NewFile`. Go marks close-on-exec every descriptor it opens itself, but a wrapped fd keeps the flags it was created with, and `forkExec` closes nothing on its own — so **every shell inherits a copy of its own master**.

That extra reference is what breaks the teardown. When f4 dies its own master closes, but the reference count never reaches zero, so the kernel never delivers `SIGHUP` to the slave side and the shell keeps running as an orphan holding its pty.

Confirmed by reading the orphaned child's descriptor table, where the inherited master is plainly visible:

```
zsh  74707  0u  CHR  16,0  /dev/ttys000
zsh  74707  1u  CHR  16,0  /dev/ttys000
zsh  74707  2u  CHR  16,0  /dev/ttys000
zsh  74707  3u  CHR  15,0  /dev/ptmx     <-- inherited master: the bug
```

A minimal reproduction that spawns a shell and then exits, mimicking f4 dying:

- without `O_CLOEXEC` → `child SURVIVED -> pty leaked (no SIGHUP)`
- with `O_CLOEXEC` → `child died -> pty correctly hung up`

## The fix

Pass `O_CLOEXEC` on both master and slave for **darwin**, **linux**, **freebsd/dragonfly** and **solaris/illumos**.

**netbsd** and **openbsd** had the same bug in a different form: they already flag `/dev/ptm`, but the descriptors `TIOCPTMGET`/`PTMGET` hand back are freshly opened and carry no `FD_CLOEXEC` of their own. Those are flagged explicitly through the existing `setCloseOnExec` helper.

This does **not** take the terminal away from the child: `Run()` installs the slave on 0, 1 and 2 with `dup2`, which clears the flag on those copies. The flag only drops the surplus inherited descriptors.

Also closes the master on the slave-open error path in `pty_unix.go`, which returned without doing so and leaked the fd — unlike every other failure path in that function.

## Tests

`pty_cloexec_test.go` covers both halves: that the flag is set, and that a child spawned through the real `Run()` path has no master in its descriptor table. **Both fail before this change** with the right diagnosis, and pass after:

```
--- FAIL: TestPTYDescriptorsAreCloseOnExec
    master descriptor lacks FD_CLOEXEC; a spawned shell would inherit it
    and keep the pty alive after f4 exits
--- FAIL: TestPTYChildDoesNotInheritMaster
    child inherited the pty master, which prevents the hangup that reaps
    it when f4 dies:
      sleep 84139 4u CHR 15,0 /dev/ptmx
```

## Verification

Builds clean for `linux`, `dragonfly`, `netbsd`, `openbsd`, `solaris`, `illumos` and `darwin`; `gofmt` clean; `TestPTY_Lifecycle` still passes.

`freebsd` fails to cross-compile, but that is **pre-existing and unrelated** — it is in a third-party dep (`go-webgpu/goffi` `fakecgo`) and reproduces on a pristine checkout of `main` without this patch. `dragonfly` shares `pty_bsd.go` and builds, so the edited code is covered.

### DIFF
--- a/pty_bsd.go
+++ b/pty_bsd.go
@@ -40,7 +40,15 @@ func ptyStep(step string, err error) error {
 }
 
 func NewPTY() (*PTY, error) {
-	masterFd, err := unix.Open("/dev/ptmx", unix.O_RDWR|unix.O_NOCTTY, 0)
+	// O_CLOEXEC matters as much as O_NOCTTY here. Go sets close-on-exec on
+	// every descriptor it opens itself, but a raw unix.Open wrapped in
+	// os.NewFile keeps whatever flags the fd was created with, and forkExec
+	// closes nothing on its own. Without the flag the shell inherits a copy
+	// of its own master, so the master never drops to zero references when
+	// f4 dies: the kernel sends no SIGHUP, the shell survives as an orphan
+	// holding its pts node, and enough restarts exhaust the pty limit until
+	// allocation fails.
+	masterFd, err := unix.Open("/dev/ptmx", unix.O_RDWR|unix.O_NOCTTY|unix.O_CLOEXEC, 0)
 	if err != nil {
 		return nil, ptyStep("open /dev/ptmx", err)
 	}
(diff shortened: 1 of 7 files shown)


unxed:
О, здорово как, у меня всё руки не доходили утечки PTY поотлавливать, а они у меня так-то довольно быстро выедают все доступные pty при прогоне тестов на локальном компе, задолбался перезагружаться.

PR #479, opened 2026-08-15
editor: map local files instead of reading them in chunks

enedzvetsky:
**Stacked on #477** — that PR's two commits are the first two here; review this one from `dd748d8`.

A local file opened in the editor is now mapped read-only and handed to the piece table as its original buffer. A mapping is one contiguous `[]byte`, so `PieceTable.View` covers the whole of an unedited file and a search scans the file itself: no chunk cache to copy out of, and no assembled snapshot to keep.

Measured on a 200 MB file of 21M lines:

| | chunk buffer (#477) | mapped |
|---|---|---|
| first search | 135 ms, 600 MB allocated | **3.3 ms, none** |
| repeated search | 3.5 ms, none (from a 200 MB cached copy) | **3.3 ms, none, no copy held** |
| touching every page | 21 ms to `read()` into the heap | **7 ms** to fault in |

Nothing is read at map time. The kernel faults pages in as they are touched and can drop them again under memory pressure without swapping, because they are clean and file-backed — so the resident cost is reclaimable in a way a heap buffer's never is.

### What still takes the old path

A remote file has no descriptor to map, an empty file has nothing to map, a legacy codepage is decoded into memory anyway, and a mapping the kernel refuses falls back with a debug line. `[Editor] MemoryMap = 0` turns it off without a rebuild.

`isLocalOSVFS` deliberately sees through wrappers around a local file system, so it cannot be the only gate; the check that actually keeps FISH+ and the cloud backends off this path is the missing file descriptor, and there is a test for it.

### Two consequences worth naming

**The indexer needed a second source.** It returned early unless there was a chunk buffer to read through, which would have left a mapped file with no line index at all. It now reads through the piece table when there is no async buffer — for the original buffer that costs nothing but the bounds arithmetic.

**SIGBUS.** A file truncated after it was mapped leaves its last pages backed by nothing, and touching one raises SIGBUS, which Go turns into a process-wide crash rather than an error an editor could report. Every goroutine that reads through the mapping — render, indexer, search, save — now sets `debug.SetPanicOnFault` and recovers, reporting the truncation once instead of taking the session down. The guard is armed only for editors that actually hold a mapping, and it re-panics on anything that is not a fault, so ordinary bugs still surface. Windows needs none of this: an open view stops the file being truncated in the first place.

### Note on open cost

`NewEditorView` rebuilds the line index at construction, and for a mapped file that scan now completes immediately rather than hitting `ErrLoading` on the second chunk — so a mapped file opens with a *complete* index instead of a growing one. On the 200 MB file that is 123 ms and a 168 MB dense index, paid at open rather than in the background. It also means the partial-index bugs (wrong line count after an early edit, a search result resolved against a half-built index) cannot arise for local files. Making that index cheap is a separate change.

### Tests

- `mapped_file_test.go` — the mapping backs the piece table without copying; empty file, absent file and a descriptorless (remote-shaped) backing all decline; truncation under a mapping is survived; the guard re-panics on a non-fault.
- `editor_mmap_test.go` — a search over a mapped editor allocates nothing and scans the mapping itself; the indexer indexes a mapped file with no async buffer; an edit lands in the add buffer, saves correctly, and the editor stays mapped across the save.

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -683,6 +683,7 @@ func runExternalEditor(pf *PanelsFrame, cmdStr, path string) {
 func showEditor(pf *PanelsFrame, v vfs.VFS, path string, f vfs.ReadAtCloser) {
 	var pt *piecetable.PieceTable
 	var buf *AsyncBuffer
+	var mapped *MappedFile
 	cpID := AppConfig.EditorDefaultCodePage
 
 	if f != nil {
@@ -697,9 +698,25 @@ func showEditor(pf *PanelsFrame, v vfs.VFS, path string, f vfs.ReadAtCloser) {
 		cpID = vfs.DetectEncoding(header, AppConfig.EditorAutodetectCodePage, AppConfig.EditorDefaultCodePage)
 
 		if cpID == 65001 {
-			buf = NewAsyncBuffer(context.Background(), f)
-			buf.prewarm()
-			pt = piecetable.NewWithBuffer(buf)
+			// A local file is mapped rather than read: the mapping is one
+			// contiguous buffer, so the piece table can hand windows onto it
+			// and a search scans the file itself instead of a copy of it.
+			// Everything else — remote, empty, or a mapping the kernel
+			// refused — keeps the lazily fetched chunk buffer.
+			if AppConfig.EditorMemoryMap {
+				var mapErr error
+				mapped, mapErr = MapEditorFile(v, f)
+				if mapErr != nil && mapErr != errNotMappable {
+					vtui.DebugLog("EDITOR: memory mapping %s failed, reading lazily instead: %v", path, mapErr)
+				}
+			}
+			if mapped != nil {
+				pt = piecetable.New(mapped.Bytes())
+			} else {
+				buf = NewAsyncBuffer(context.Background(), f)
+				buf.prewarm()
+				pt = piecetable.NewWithBuffer(buf)
+			}
 		} else {
 			fullData := make([]byte, size)
 			_, _ = f.ReadAt(context.Background(), fullData, 0)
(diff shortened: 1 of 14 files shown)


unxed:
1. Это нам удалённое индексирование в FISH+ не сломает ведь?

UPD: а, вижу:

> isLocalOSVFS deliberately sees through wrappers around a local file system, so it cannot be the only gate; the check that actually keeps FISH+ and the cloud backends off this path is the missing file descriptor, and there is a test for it.

2. Тест падает:

```
2026-08-15T08:08:13.4388782Z === RUN   TestMappedEditor_EditsAndSaves
2026-08-15T08:08:13.4389480Z     editor_mmap_test.go:159: content = "XYZ XYZ o world\n", want "XYZ hello world\n"
2026-08-15T08:08:13.4390071Z --- FAIL: TestMappedEditor_EditsAndSaves (0.00s)
```

enedzvetsky:
The CI failure was a real bug, not a flaky test, and it was hiding underneath the chunk cache until mapping exposed it.

`OSVFS.PatchInPlace` validated its pieces *as it wrote them*. In-place patching can only express edits that leave every unchanged piece at the offset it already occupies, so an insertion has to be refused — but the refusal came after the pieces before it had already gone to disk. Typing `XYZ ` at the start of `hello world` wrote those four bytes over `hell`, then refused the patch. The editor's save then fell back to a full rewrite, which reads the unchanged pieces back through its own buffer — for a mapped file, the file itself — and saved `XYZ XYZ o world`.

It reproduces without the editor at all:

```
--- FAIL: TestPatchInPlace_RejectsWithoutWriting
    file = "XYZ o world\n" after a refused patch, want it untouched as "hello world\n"
```

macOS hid it because a private mapping there did not show writes made through another descriptor; on Linux it does, which is why only CI failed. A buffer that had cached the region survived it too — which is why this has been latent rather than visible.

`bf5d6a4` adds `ValidateInPlacePieces`, run over the whole plan before the file is opened, so a refusal leaves it untouched. Applied to `DisksVFS.PatchInPlace` as well, where writing before validating is aimed at a raw device.

PR #480, opened 2026-08-15
piecetable: hold line offsets in blocks, relative to a base

enedzvetsky:
**Stacked on #479** (which is stacked on #477) — review this one from `36180f7`.

The line index was one `int` per line in a single slice, grown by `append`. On a 200 MB file of 21M lines that is 168 MB of offsets sitting in 330 MB of heap once the doubling slack is counted, and every edit walked all of it: typing one character at the top shifted 21M offsets, and pressing Enter reallocated the array to splice one value in.

Offsets now live in blocks of a few thousand, each stored as a distance from the offset its block begins at. That buys three things at once:

- **four bytes per line instead of eight**, and filling the index allocates another block rather than copying the whole array;
- **an edit that adds no lines** moves everything after it by adjusting one base per block — thousands of values instead of millions;
- **an edit that does add lines** splices into a single block, leaving the blocks after it to learn only that their lines are numbered differently.

Measured against the old implementation at 21M lines:

| | dense `[]int` | blocked |
|---|---|---|
| build | 95 ms | **44 ms** |
| heap | 330 MB | **80 MB** |
| typing at line 1 | 5.66 ms per keystroke | **4.8 µs** |
| Enter at line 1 | 32.6 ms | **9.8 µs** |

Lookups stay logarithmic — finding the block and then the line inside it is two short searches rather than one over every line in the file.

### The 4 GB fallback

A file past 4 GB cannot describe its offsets in 32 bits relative to anything, so it falls back to the plain `[]int` this used to be for every file. The fallback is reached by construction rather than by a size check: an index that meets an offset it cannot encode switches layout and carries on, which also covers a pathological block whose lines span more than 4 GB between them.

### Why you can trust the rewrite

The exported API and its behaviour are untouched — `GetLineOffset`, `LineCount`, `GetLineAtOffset`, `AppendOffsets`, `Rebuild`, `UpdateAfterInsert`, `UpdateAfterDelete` all answer exactly as before, so none of the ~135 call sites in the editor change.

`lineindex_equivalence_test.go` carries a copy of the old dense implementation as an oracle and drives both through the same random stream of inserts, deletes and newline splices, comparing every line offset and probing the reverse mapping at the ends, the block boundaries and a random sample after every step. Alongside it: a test that walks lines across several block boundaries, one that deletes whole blocks at once, and one that forces the 4 GB fallback and keeps editing afterwards.

### What this sets up

With the index no longer the expensive part, the remaining work from the same plan is the index state machine and progress reporting, and then the remote residency policy — both of which wanted a cheap index first.

### DIFF
--- a/piecetable/lineindex.go
+++ b/piecetable/lineindex.go
@@ -3,16 +3,305 @@ package piecetable
 import "sort"
 import "sync"
 
-// LineIndex stores start offsets of each line.
+const (
+	// lineBlockTarget is how many line offsets a block holds when it is filled
+	// or split. Small enough that rewriting one is cheap, large enough that a
+	// file of millions of lines has thousands of blocks rather than millions.
+	lineBlockTarget = 4096
+	// lineBlockMax is the size a block is split at, so that repeated inserts
+	// into the same place cannot grow one block without bound.
+	lineBlockMax = 2 * lineBlockTarget
+
+	// maxRelativeOffset is what fits in the uint32 an entry is stored as.
+	maxRelativeOffset = 1<<32 - 1
+)
+
+// fitsRelativeOffset reports whether a distance from a block's base can be
+// stored in the uint32 an entry is. The comparison goes through uint64 because
+// the limit does not fit in an int on a 32-bit build, where it is also
+// unreachable: no offset there can exceed 2 GB in the first place.
+func fitsRelativeOffset(rel int) bool {
+	return rel >= 0 && uint64(rel) <= maxRelativeOffset
+}
+
+// lineBlock holds the offsets of a run of consecutive lines. Each offset is
+// stored as a distance from base — the absolute offset of the block's first
+// line — so entries[0] is always zero and an edit before the block moves all
+// of it by moving base alone.
+type lineBlock struct {
+	firstLine int
+	base      int
+	entries   []uint32
+}
+
+// LineIndex stores the byte offset each line starts at.
+//
+// The offsets live in blocks of a few thousand, each entry held relative to the
+// offset its block begins at. Three things follow, and all of them matter on a
+// file with millions of lines:
+//
+//   - four bytes per line instead of eight, and filling the index allocates one
+//     more block instead of reallocating and copying the whole array — 21M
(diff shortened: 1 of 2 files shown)


unxed:
Класс! Только сборка на arm пока падает:
```
Error: piecetable/lineindex.go:104:63: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: piecetable/lineindex.go:175:24: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: piecetable/lineindex.go:213:19: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: Process completed with exit code 1.
```

unxed:
Всё ещё падает сборка на arm:
```
# github.com/unxed/f4/piecetable
Error: piecetable/lineindex.go:104:63: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: piecetable/lineindex.go:175:24: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: piecetable/lineindex.go:213:19: maxRelativeOffset (untyped int constant 4294967295) overflows int
Error: Process completed with exit code 1.
```

enedzvetsky:
Fixed — the 32-bit build was two separate constant-overflow problems, both mine.

`maxRelativeOffset` is 4294967295, which does not fit in an `int` where `int` is 32 bits, so comparing an `int` against it does not compile. It is now behind a helper that goes through `uint64`:

```go
func fitsRelativeOffset(rel int) bool {
	return rel >= 0 && uint64(rel) <= maxRelativeOffset
}
```

On a 32-bit build the limit is unreachable anyway — no offset there can exceed 2 GB — so the check simply never fires.

The test for the 4 GB fallback had the same problem in a form the compiler catches even harder: `int(huge)` where `huge` is a typed `int64` constant of 2^33 is a constant conversion, which is a compile error rather than a runtime skip. It now builds the value at runtime and skips when `int` cannot hold it.

Verified with `GOARCH=386` and `GOARCH=arm` (`go vet`, which compiles the tests too) alongside the native run.

PR #484, opened 2026-08-15
macos: stamp the icon onto the binary instead of setting it at runtime

enedzvetsky:
#456 set the Dock tile with `[NSApp setApplicationIconImage:]`, the documented runtime override for a process that has no app bundle. On macOS 26 it turns out to do nothing for exactly that case: the Dock tile and `NSRunningApplication.icon` keep returning the generic green executable icon, whether the call happens before or after `finishLaunching`, and whether or not a bundle is present.

`[NSWorkspace setIcon:forFile:options:]` does work. Stamping the icon onto the executable file makes Launch Services pick it up immediately — the running process's tile updates in place — and Finder shows the same icon for the binary from then on. The stamp lives in the file's `com.apple.FinderInfo` / `com.apple.ResourceFork` extended attributes; it does not touch the executable's contents, so the code signature stays valid.

### When it does nothing

- **Running from F4.app** — the bundle's `CFBundleIconFile` already covers Dock and Finder, so the path returns early after resolving symlinks and matching `.app/Contents/MacOS/`.
- **Already stamped** — `com.apple.FinderInfo` bit `kHasCustomIcon` is checked first, so only the very first bare-binary launch writes anything. Rewriting a resource fork on every start would be pure disk churn.
- **Read-only location** — `/usr/local/bin` without write access, a sealed volume, and so on. `setIcon:forFile:` returns NO, that goes to the debug log, and the tile stays generic. Nothing fails.
- **X11/Wayland backends** (XQuartz) — unchanged, AppKit is left alone.

The embedded asset moves from `f4-512.png` to the already-present `f4.icns`, so every size the Dock and Finder ask for comes out of one stamp instead of being downsampled from a single 512px bitmap. `NSImage` still owns the decode; the `alloc`/`init` pairing (rather than the autoreleasing convenience constructors) is kept, since this still runs before gogpu sets up an autorelease pool.

Tested on macOS 26 with a bare `go build` binary launched from a terminal: first run stamps and the Dock tile turns into the F4 icon while the process is running; subsequent runs skip the write; running the same build from `F4.app` is unaffected.

### DIFF
--- a/window_icon_darwin.go
+++ b/window_icon_darwin.go
@@ -4,23 +4,41 @@ package main
 
 import (
 	_ "embed"
+	"os"
+	"path/filepath"
+	"strings"
 	"unsafe"
 
 	"github.com/ebitengine/purego"
 	"github.com/ebitengine/purego/objc"
 	"github.com/unxed/vtui"
+	"golang.org/x/sys/unix"
 )
 
-//go:embed assets/icon/generated/f4-512.png
-var darwinDockIconPNG []byte
+//go:embed assets/icon/generated/f4.icns
+var darwinIconICNS []byte
 
-// applyDarwinDockIcon gives the Dock tile an image. gogpu switches the
-// process to NSApplicationActivationPolicyRegular, which is what makes the
-// tile appear at all, but the tile's image comes from the app bundle's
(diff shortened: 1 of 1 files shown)


unxed:
Тема! Спасибо :)

PR #486, opened 2026-08-15
editor: let a file system search its own copy

enedzvetsky:
**Stacked on #485** — review this one from `d314090`.

Searching a remote file meant downloading it. The editor reads the whole buffer to scan it, and for a file on a FISH+ host that read is a transfer — of the entire file, to find a string the host could have found itself. `vfs.Search` has been there for a while and the viewer has used it since it was added; the editor never has.

### When the search is handed over

Four things have to hold:

| condition | why |
|---|---|
| the file system offers `HasSearch` | nothing to ask otherwise |
| the buffer is still the file on disk | the offsets come back as positions in what is stored; an edit moves the text away from that |
| the content is raw UTF-8 | a decoded buffer's offsets are not the file's either |
| no regex, no whole-word | not what a fixed-string grep answers |

The "still the file" test is `pt.GetState().Equals(cleanState)`, which is exactly the question already used to decide whether the buffer is modified.

### Confirming what comes back

FISH+ folds case and matches fixed strings, so its answer is a *superset* of some searches. Every candidate is confirmed against the buffer's own rules before it is used — reading a few bytes at that offset rather than the file — which is what makes a case-sensitive search over a case-folding host correct. A folded match can also cover a different number of bytes than the pattern (`K` U+212A matches `k`), so the length is measured there rather than assumed.

### The part that matters most

A host-side search caps how many matches it reports — FISH+ stops at ten thousand. So a list can be short of the truth, and finding nothing usable in it **cannot** mean "not in the file". It falls back to scanning the buffer, which is slower and certain. Every uncertainty resolves the same way: no answer, an error, or an unconfirmable candidate all end up on the local scan. The only thing delegation can do is answer faster, never differently.

There is a test for exactly this: a stub host that reports only the first three of fifty matches, and a search past them that must decline rather than claim not-found.

### Tests

`editor_search_remote_test.go` — a searchable file system stub with the host's semantics (fixed, folded, capped): delegation finds and confirms a match; a case-sensitive search rejects the folded candidate and lands on the right one; forward, backward and next all pick the correct occurrence; a truncated answer falls back; and delegation declines for regex, whole-word, an edited buffer, a decoded codepage, and a file system without the capability.

### What is left of the remote work

The other half of the plan's remote stage is a windowed residency policy — bounding what a remote file keeps resident instead of accumulating every chunk it has ever read. With host-side line indexing (already in `main`) and this, a large remote file no longer has to travel to be opened, indexed or searched, so what remains is memory rather than transfer.

### DIFF
--- a/editor_index_status.go
+++ b/editor_index_status.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"time"
+
+	"github.com/unxed/vtui"
+)
+
+// IndexPhase is where the line index stands. The editor can work with an index
+// that is still filling — it just cannot answer questions about lines it has
+// not reached yet — so the phase is what tells the difference between "there is
+// no more to find" and "there is more, we are looking".
+type IndexPhase int
+
+const (
+	// IndexIdle: nothing has been scanned and nothing is scanning.
+	IndexIdle IndexPhase = iota
+	// IndexScanning: a scan is running and the index is still growing.
+	IndexScanning
+	// IndexComplete: the index describes every line of the buffer.
+	IndexComplete
+	// IndexFailed: the scan stopped on an error and the index is short.
+	IndexFailed
+)
+
+func (p IndexPhase) String() string {
+	switch p {
+	case IndexScanning:
+		return "scanning"
+	case IndexComplete:
+		return "complete"
+	case IndexFailed:
+		return "failed"
+	default:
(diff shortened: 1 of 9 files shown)


unxed:
О, это я собирался сделать и забыл, спасибо огромное!

PR #538, opened 2026-08-20
pty sync: replace POSIX '# f4_sync' comment with '&& true f4_sync'

refaim:
## Problem

Navigating panels with a background PTY open spams the terminal on stock zsh:

```
hrt@localhost ~ %  cd '/Users/hrt' # f4_sync
cd: too many arguments
```

The cwd sync line ` cd '<path>' # f4_sync` relies on `#` starting a comment, but stock zsh ships with `interactive_comments` off (oh-my-zsh enables it, plain macOS zsh does not). zsh passes `#` and `f4_sync` to `cd` as extra arguments, so every panel navigation prints `cd: too many arguments` and the shell cwd never follows the panel. The error line also defeats the ANSI parser's excision, leaving junk in the scrollback.

## Fix

Send the marker as an executable no-op instead of a comment:

```
 cd '<path>' && true f4_sync
```

`true` ignores its arguments in every POSIX-ish shell including fish, so the line works regardless of `interactive_comments`. Changed in both generators (`panels_frame.go`, `plugins/netfox/fish_vfs.go`) and in the parser's excision needle (`ansi_parser.go`). The Windows `& rem f4_sync` variant is untouched.

Also made the Unix excision test meaningful: it asserted `strings.Contains(log, "")` which is vacuously true; it now asserts the marker never leaks into the visible log.

### DIFF
--- a/ansi_parser.go
+++ b/ansi_parser.go
@@ -79,9 +79,9 @@ func (p *AnsiParser) Process(data []byte) {
 		if startIdx == -1 {
 			break
 		}
-		endIdx := bytes.Index(data[startIdx:], []byte("' # f4_sync"))
+		endIdx := bytes.Index(data[startIdx:], []byte("' && true f4_sync"))
 		if endIdx != -1 {
-			actualEnd := startIdx + endIdx + len("' # f4_sync")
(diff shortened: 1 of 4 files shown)


unxed:
Great (and long awaited) fix, thanks!

PR #547, opened 2026-08-20
ci: test natively on 6 os/arch cells; make the timing-sensitive tests deterministic

refaim:
Follow-up to #545, three commits. Depends on nothing: unxed/vtui#44 is already merged and this branch requires the tag that carries it (`vtui v0.1.234`).

## 1. Test matrix: 3 → 6 native cells

`ubuntu-24.04-arm`, `windows-11-arm` and `macos-15-intel` join the test matrix, adding native linux/arm64, windows/arm64 and darwin/amd64 coverage. All three are free hosted runners at native speed — linux/arm64 is actually the fastest cell of the six (~2 min). No BSD/Solaris cells: GitHub has no such runners, and QEMU-based VMs would be slow and flaky enough to hurt more than help.

## 2. Highlight budget tests: deterministic instead of wall-clock

`TestEditor_HighlightSlice_StopsOnBudget` and `TestEditor_BackgroundWalker_NoLongUIStalls` measured real elapsed time and asserted bounds on it — so any stall of a shared runner (GC, antivirus, noisy neighbor) failed them. `StopsOnBudget` failed exactly that way on this repo's own CI ("slice occupied the UI thread for 101.948ms, budget is 4ms" — that's the machine, not the code), and the walker test did the same on windows-11-arm.

The slice budget clock now sits behind a seam (`hlNow`), and the tests install a fake clock advanced by the mock highlighter's per-line cost. Every bound became an exact equality: a slice stops at exactly the first clock-stride multiple that exhausts the budget, and one walker task may grow the chain by exactly that many lines. Verified with `-race -count=10`.

## 3. Deflake: archive progress and clipboard

**plugins/archive**: the seeking-progress test slept a fixed 400ms and asserted that a 250ms ticker had already fired — a bet the scheduler loses on a loaded runner (flaked on ubuntu-latest). It now waits for the status to arrive, and the mock reporter gets the mutex its cross-goroutine use always needed. The Close-deadlock reproducer's verdict window widens 500ms → 10s (happy path returns instantly; the guarded regression — Close blocking forever — is still caught).

**cmd/f4**: clipboard tests ran against the machine's real clipboard — on macOS that's up to ~400 pbpaste spawns inside one test's 2s deadline (`TestAction_PanelCopySelectedNames` flaked on macos-latest), and locally it clobbers the developer's actual clipboard. `TestMain` now keeps clipboard traffic in vtui's process-local buffer via `vtui.SkipOSClipboard` (unxed/vtui#44, in v0.1.234) plus `DisableTerminalClipboard()` — which also stops the OSC 52 fallback from spraying base64 over the test logs. `TestTerminalView_ProcessFar2lInteract_ConcurrentRace` opts back out for its own scope: hammering the real OS clipboard is the point of that reproducer.

## Verification

Fork CI on this exact head: all 6 test cells green, all build cells green (including windows7 legacy). The previously-flaky tests were additionally hammered locally with `-race -count=10`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -330,13 +330,21 @@ jobs:
 
   # Эта джоба создает постоянные релизы при пуше тэгов v*.*.*
   test:
-    name: Test (${{ matrix.os }})
+    name: Test (${{ matrix.target }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        # One hosted runner per native os/arch cell, labeled to match the
+        # build matrix.
+        include:
+          - { os: ubuntu-latest,    target: linux/amd64 }
+          - { os: ubuntu-24.04-arm, target: linux/arm64 }
+          - { os: macos-latest,     target: darwin/arm64 }
+          - { os: macos-15-intel,   target: darwin/amd64 }
+          - { os: windows-latest,   target: windows/amd64 }
+          - { os: windows-11-arm,   target: windows/arm64 }
     env:
       CGO_ENABLED: 0
     steps:
--- a/cmd/f4/editor_highlight_budget_test.go
+++ b/cmd/f4/editor_highlight_budget_test.go
@@ -10,18 +10,48 @@ import (
 	"github.com/unxed/vtui"
 )
 
-// mockSlowHighlighter spends a fixed amount of time per line, standing in for
-// a parser whose per-line cost is high enough that a fixed batch size turns
-// into a visible freeze (Colorer through wasm, in practice).
+// fakeHLClock replaces the slice budget clock, so a "slow" highlighter costs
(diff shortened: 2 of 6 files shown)


refaim:
One more thing worth doing on the repository-settings side (only you can, it's not in the workflow files): now that the Test jobs run on real PRs, consider marking them as **required status checks** for merging into main — Settings → Branches (or Rulesets) → branch protection for `main` → Require status checks to pass → add the six `Test (...)` checks. That way a red test job actually blocks the merge button instead of just coloring the PR, and flaky-looking failures get investigated rather than merged past. The Build jobs can stay optional or be added too, at your discretion.

unxed:
Не-не, блокировать приём PR мы не будем, потому что я бывает руками за человеком следующим же коммитом мелочи правлю. Правильно принять изменения новичка, чтобы мотивировать его присылать ещё, и поправить за ним по мелочи, а не давать ему от ворот поворот "иди выучи всё программирование сразу". А опытные и сами с таким разберутся.

PR #550, opened 2026-08-20
Guard GUI backend initialization panics

Zoinen:
## Problem

F4's automatic GUI selection already treats a backend error as recoverable and tries the next available backend. A panic during backend initialization bypasses that mechanism and terminates the process before fallback can happen.

We observed this on Linux/ARM64 with the Wayland backend. The cursor decoder in `github.com/neurlang/wayland` panicked inside `window.DisplayCreate`, before F4's UI setup callback ran, so launching `f4 --gui` appeared to do nothing except create a crash report.

The underlying ARM64 swizzle issue is addressed separately in [neurlang/wayland#28](https://github.com/neurlang/wayland/pull/28).

## Fix

Convert a panic into a backend startup error only while the native backend is initializing. This lets `tryRunDefaultGui` record the Wayland failure and continue to another available backend.

The recovery boundary is deliberately narrow:

- startup is marked complete only after `SetupUI` and `openDashEFileIfRequested` return;
- a panic after that point is re-thrown, preserving F4's normal crash reporting for runtime/application faults;
- external `qt` and `ext:` backends keep their existing execution path and are not wrapped;
- successful GUI startup and normal backend errors are unchanged.

An atomic startup flag avoids assumptions about which goroutine invokes the backend setup callback.

## Tests

Added focused tests proving that:

- a pre-setup backend panic becomes a descriptive error containing the backend name;
- a post-setup panic is not suppressed.

Validated with:

```text
go test -run '^TestRunGui(ConvertsStartupPanicToError|PreservesPanicAfterStartup)$' ./cmd/f4
go build ./cmd/f4
```

For integration validation, this branch was built in a Go workspace with the Wayland fix from `neurlang/wayland#28` on Linux/ARM64. Running the resulting binary with `--gui=wayland` initialized `BACKEND: active rendering backend is "wayland"` and remained running without a crash.

This PR intentionally contains no dependency replacement or reference to a personal fork.

### DIFF
--- a/cmd/f4/gui_unix.go
+++ b/cmd/f4/gui_unix.go
@@ -3,18 +3,42 @@
 package main
 
 import (
+	"fmt"
 	"strings"
+	"sync/atomic"
 
 	"github.com/unxed/vtui"
 )
+
+func runGuiWithStartupRecovery(backend string, startupComplete *atomic.Bool, run func() error) (err error) {
+	// A display backend is an optional integration and may panic while
+	// initializing (for example, while loading a Wayland cursor theme). Treat
+	// that as a backend failure so automatic GUI selection can try another
+	// available backend. Panics after SetupUI completes are application faults
+	// and must retain the normal crash-reporting path.
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			if startupComplete.Load() {
+				panic(recovered)
+			}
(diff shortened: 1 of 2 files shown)


unxed:
О, дело. Давно руки не доходили

Issue #523, opened 2026-08-18
Падение f4 при работе с архивами


Alexey5112:
<img width="1483" height="787" alt="Image" src="https://github.com/user-attachments/assets/cc409525-bf8b-494e-847d-0b91145a90c5" />
жмём энтер, входим в архив:

<img width="1483" height="787" alt="Image" src="https://github.com/user-attachments/assets/1b83e2cc-54fa-4d99-9227-6c3134b91c0c" />
жмём энтер (архив же) - закрывается  f4 молча
лог:

[crash_20260818_210014_13646.log](https://github.com/user-attachments/files/31192225/crash_20260818_210014_13646.log)

Alexey5112:
<img width="1080" height="1279" alt="Image" src="https://github.com/user-attachments/assets/7fc3d133-7ba7-4cc9-8b0d-3a0c28b9e892" />
Так же падение при попытке зайти в файл *.zip нулевой длины, похоже причина та же.

Zeroes1:
а сейчас? (буду косплеить Ивана :)

Alexey5112:
<img width="882" height="153" alt="Image" src="https://github.com/user-attachments/assets/69e57ba0-6626-449c-b231-ac31821f8230" />
на нулевом zip не падает

<img width="444" height="283" alt="Image" src="https://github.com/user-attachments/assets/257e23ae-2b6c-4fb2-813c-8cbcab856264" />

на вложенном архиве не падает

<img width="589" height="366" alt="Image" src="https://github.com/user-attachments/assets/54b745df-5065-4d84-8e94-3e0aad725f3b" />
но кажется, неправильное поведение всё равно, почему в архиве 

history.log.1.gz лежит файл history.log.1.gz ???

unxed:
~/4$ ./zipper x ./history.log.1.gz
Error: format archives.Gz does not support extraction

О как. Изучим!

([zipper](https://github.com/unxed/zipper) если что это либа и cli-тулза, на базе которой в f4 архивы работают)

Zoinen:
I'm zoin-bot, investigation completed on Windows 10 x64.

The original zero-byte archive crash is already fixed upstream by commit https://github.com/unxed/f4/commit/6ca1b558415178d9682576bb18a666587a92bf5f, so no duplicate f4 PR is needed.

[...]

PR #551, opened 2026-08-20
fix: crash opening 0-byte archives via typed nil VFS

Zeroes1:
close: fix all or part https://github.com/unxed/f4/issues/523

## Problem
Entering an empty/0-byte .zip file crashes f4 with
`runtime error: invalid memory address or nil pointer dereference`
in `(*ArchiveVFS).Close` (vfs.go:1266, `v.mu.Lock()`).

## Root cause
`NewArchiveVFSContext` returns `(*ArchiveVFS)(nil), err` on failure
(e.g. `archive.OpenFS` rejects a 0-byte file). `ArchiveProvider.Open`
returned that value directly, so the typed nil got boxed into the
`vfs.VFS` interface as a NON-nil value. The async panel open
(`openVFSAsync`) then ran `if newVFS != nil { _ = newVFS.Close() }`,
and calling a method on the nil pointer panicked.

Any provider can leak a typed nil this way, so the fix is defensive
at the panel layer in addition to the archive source.

## Fix
- `plugins/archive/provider.go`: `Open` returns an untyped `nil`
  interface on error, so `newVFS == nil` is true again.
- `cmd/f4/file_panel.go`: added `isNilVFS` (reflect-based, detects a
  typed nil inside a non-nil interface); `openVFSAsync` uses it for
  every `Close()` guard and normalizes a typed nil into a plain error
  before the success/error branches. This protects all providers.
- Error dialog wording: archive failures now show
  "Open Error / Failed to open ..." instead of the
  network-oriented "Connection Error / Failed to connect ...".
  Only the `zipper/archive` provider is affected; other VFS plugins
  keep the existing wording.

## Testing
Entering a 0-byte `.zip` no longer crashes; the panel shows
"Open Error" and falls back to the previous directory. Regular
archives still open normally.

### DIFF
--- a/cmd/f4/file_panel.go
+++ b/cmd/f4/file_panel.go
@@ -580,6 +580,28 @@ func sameVFSInstance(a, b vfs.VFS) bool {
 	return a == b
 }
 
+// isNilVFS reports whether v is nil, including a typed nil wrapped inside a
+// non-nil interface (e.g. (*ArchiveVFS)(nil) returned as vfs.VFS). Such a
+// value compares != nil but dereferences to a panic on any method call.
+func isNilVFS(v vfs.VFS) bool {
+	if v == nil {
+		return true
+	}
+	rv := reflect.ValueOf(v)
+	switch rv.Kind() {
+	case reflect.Ptr, reflect.Map, reflect.Slice, reflect.Func, reflect.Chan:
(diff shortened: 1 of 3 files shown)


unxed:
О, ух ты, круто!

PR #554, opened 2026-08-21
highlight: the Executable rule attribute must not match directories

### LINKED ISSUE #419: Подсветка файлов с атрибутом Executable также подсвечивает папки
Добавление правила с этим атрибутов подсвечивает все папки в Linux (если только правило для папок не указано перед ним)

refaim:
Fixes #419.

On Unix every directory carries the x bit — it grants traversal, not execution — so a highlight rule with the Executable attribute lit up every folder, unless a Directory rule happened to sort before it (the workaround #419 mentions; the default themes dodge the same trap with an explicit `ExcludeAttributes = Directory` on their script rules).

The fix interprets the rule attribute as what it names — an executable *program* — at the one place all VFS backends route through: `HighlightRule.Match` now treats a directory as not-executable regardless of its mode bits. The complement becomes consistent too: an `ExcludeAttributes = Executable` rule now matches directories instead of skipping them.

`VFSItem.IsExecutable` itself intentionally keeps the raw bit: its other consumers (fusefs mode synthesis, terminal runnability) already branch on `IsDir` first and want the honest value.

Covered by new table cases in `TestHighlightRule_MatchAttributes` (fail without the fix); verified manually on macOS with an `IncludeAttributes = Executable` rule — folders stay uncolored, `+x` files light up.

### DIFF
--- a/cmd/f4/highlight_files.go
+++ b/cmd/f4/highlight_files.go
@@ -258,7 +258,11 @@ func (r *HighlightRule) Match(item *vfs.VFSItem) bool {
 		case AttrHidden:
 			return item.IsHidden == set
 		case AttrExecutable:
-			return item.IsExecutable == set
+			// On Unix every directory carries the x bit (it means "may be
+			// entered", not "may be run"), so a bare IsExecutable check would
+			// light up all folders (#419). The rule attribute means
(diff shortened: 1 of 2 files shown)


unxed:
Great, thanks!

Issue #561, opened 2026-08-21
размер диалогов настроек


Alexey5112:
f4-gui linux, версия:
<img width="783" height="128" alt="Image" src="https://github.com/user-attachments/assets/617fba6b-b509-43b0-ae26-ee891e31ed2a" />
при окне ф4 меньше высоты диалога настроек окно обрезается, отсутствует скролл, не видно кнопок ok/cancel

<img width="1672" height="743" alt="Image" src="https://github.com/user-attachments/assets/7f233ff0-d2a6-4eb5-b27c-ec2b3874a5db" />

при уменьшении окна ф4 с открытым диалогом настроек периодически не проходит рефреш содержимого, наблюдаем чёрный прямоугольник:

<img width="1672" height="743" alt="Image" src="https://github.com/user-attachments/assets/1a89ef02-1c7b-47f3-8fc6-25f9b8898374" />

случайное нажатие в область окна настроек - частичное отрисовывание: 

<img width="1608" height="734" alt="Image" src="https://github.com/user-attachments/assets/158c060f-b4b1-4b50-917e-ba2130f849e4" />

при нажатии Esc (или другом способе закрытия окна настроек) сразу всё отрисовывает

Zoinen:
Fixed and verified on native Windows 11 with the Win32 GUI. The settings dialog now stays inside a short viewport; `Page Down` reaches `OK`/`Cancel`, the mouse wheel scrolls even over child controls, and repeated native resize cycles redraw without artifacts.

Pull request: https://github.com/unxed/f4/pull/567

Zoinen:
I, zoin-bot, completed verification of issue #561 on Linux aarch64 (Fedora Asahi Remix 44, KDE Plasma Wayland).

I inspected all four attached screenshots and verified that the original fix from f4 PR #567 is present in current upstream: both settings dialogs stay within short viewports, PageDown and mouse scrolling reach the action buttons, and repeated resize/click/Escape cycles did not reproduce a persistent black or partial redraw on the rebuilt ARM64 X11 binary.

The live test also exposed a residual shared edge case: after scrolling focus to a bottom control, shrinking the terminal could leave that focused control outside the new viewport. I fixed this in vtui by reapplying the existing focus-visibility invariant after modal resize and added a regression test. Pull request: https://github.com/unxed/vtui/pull/106

[...]

PR #567, opened 2026-08-21
fix: keep settings dialogs accessible in short windows

Zoinen:
## Summary
- add a regression test for Panel Settings in a short 100×30 viewport
- keep full-layout validation separate from short-viewport scrolling behavior
- document the three considered solutions for issue #561

The implementation is included in `github.com/unxed/vtui v0.1.255`, which is now consumed by the current f4 `main`. This PR adds the f4-level regression coverage and documentation on top of that released fix.

## Verification
- Native Windows 11 / Win32 GUI: opened Panel Settings in a 30-row viewport, navigated to `OK`/`Cancel` with Page Down, scrolled via the wheel while pointer was over a child control, and ran repeated window-resize cycles without redraw artifacts.
- `go test ./cmd/f4 -count=1`
- `go build -o build\\f4-issue561-rebased-gui.exe ./cmd/f4`

### DIFF
--- a/cmd/f4/issue561_test.go
+++ b/cmd/f4/issue561_test.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/unxed/vtinput"
+	"github.com/unxed/vtui"
+)
+
+type issue561ViewportRenderer struct {
+}
+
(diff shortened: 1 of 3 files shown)


unxed:
Тут надо конфликты поправить

Issue #632, opened 2026-08-22
Интернационализация ввода


unxed:
У нас уже сделан вывод на сложных языках вроде языков, где нужен BiDi и бенгальского с его составными символами. Но вот ввод на них ещё никто не делал. Надо сделать. И в диалогах и в редакторе и в командной строке и где ещё у нас текст вводится.

Zoinen:
I'm zoin-bot, running on Windows 10 x64. Work on issue #632 is complete.

Implemented and merged:
- f4 #644: full BiDi/grapheme-aware editor input and caret handling across the editor and command-line paths.
- f4 #646, #649, and #650: CI-driven fixes for Indic/Divehi grapheme boundaries and long-line performance. Unwrapped-line visual clusters and prefix widths are now cached, so repeated cursor movement no longer rebuilds a 100 KB line.
- f4 #652: mapped-file fault protection for the fully-read indexing goroutine; its Windows ARM64 stack fault is now recovered through the same guard as lazy indexing.
- vtui #70: MultiLineEdit BiDi/grapheme input support, merged and included in vtui v0.1.263.

Validation:
- Native Windows build and --version launch smoke test passed.
- Targeted Unicode, grapheme, BiDi, wrap-safety, textlayout, and long-line performance tests passed locally.
- The final f4 CI matrix for #652 passed 24/24 checks, including Windows ARM64, Linux amd64/arm64, race detector, and all build targets.

Merged PRs: https://github.com/unxed/f4/pull/644, https://github.com/unxed/f4/pull/646, https://github.com/unxed/f4/pull/649, https://github.com/unxed/f4/pull/650, https://github.com/unxed/f4/pull/652

— zoin-bot

PR #644, opened 2026-08-22
Support grapheme-aware and BiDi text input (#632)

Zoinen:
## Summary

- enable vtui's full visual caret mode at the f4 application boundary
- make editor and command-line insertion, navigation, backspace, delete, wrapping, selection, and rendering grapheme-aware and BiDi-aware
- add regression coverage for Hebrew visual caret movement, combining marks, Bengali/Indic wrapping, and editor rendering

## Validation

- targeted go test suites pass for textlayout and the new f4 Unicode input tests
- native Windows x64 build succeeds and launches against a real terminal session (--version and interactive launch/exit smoke test)
- the full cmd/f4 suite currently hits an existing Windows memory fault in scanEditorWrapSafety, which is outside this diff

## Dependency

This f4 change uses the existing vtui APIs and requires the upstream grapheme/BiDi multiline-input implementation from [unxed/vtui#70](https://github.com/unxed/vtui/pull/70). That PR is open, mergeable, and its CI is green.

— zoin-bot

### DIFF
--- a/cmd/f4/editor_view.go
+++ b/cmd/f4/editor_view.go
@@ -1582,9 +1582,8 @@ func (ev *EditorView) DisplayObject(scr *vtui.ScreenBuf) {
 			}
 			runesProcessedInLine += fragRuneCount
 
-			_, startVCol := ev.engine.LogicalToVisual(frag.ByteOffsetStart)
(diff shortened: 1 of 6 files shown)


unxed:
CI падает

PR #676, opened 2026-08-22
Fix Auto Update dialog missed first click

### LINKED ISSUE #666: проблемы с мышью, миссклик
f4-gui linux kubuntu 26.04:

Image

курсор на левой панели, в диалоге автообновление, при клике на "проверить сейчас" нет никакой визуальной реакции на клик мышкой, автоматически кликаю второй раз (не дабл клик, а два клика подряд, окно не перерисовано)

Image

тут идёт смена диалога, и выделение перемещается, в данном случае под диалог:
Image

Image

Zoinen:
Closes #666

## Root cause

The Auto Update settings dialog closed on `Check now`, but its click handler then ran the manual GitHub request synchronously on the UI event-dispatch goroutine. Until the request returned, f4 could not redraw the closed dialog or process the next input event, which made the first mouse click appear lost and left the underlying panel in a stale visual state.

## Fix

The manual check now runs in a background goroutine after the dialog closes. Its existing result/error delivery still goes through `FrameManager.PostTask`, so UI changes remain on the UI thread. The same issue was audited across update-check call sites; startup checks were already asynchronous.

## Regression coverage

Added `TestActionUpdateSettings_ManualCheckDoesNotBlockMouseDispatch`, which holds an HTTP response open and verifies that the dialog's mouse dispatch returns immediately while the request continues and completes in the background.

## Validation

- `go test ./... -count=1`
- `go build -o C:\Temp\f4-666.exe .\cmd\f4`
- Native Windows `f4-666.exe --version`, `--wine-probe`, and `-test-plugins`
- Native Windows `--gui=win32` smoke test: opened the Auto Update settings through the command palette, activated `Check now`, observed immediate frame removal/redraw and a background `UPDATER: Checking for updates` log entry.

The maximized test display exposed a host-only logical/physical size mismatch (160x49 logical cells in a 1440x853 capture), so the dialog was partly outside the screenshot; the native backend still accepted and processed the event path successfully.

— zoin-bot

### DIFF
--- a/cmd/f4/actions.go
+++ b/cmd/f4/actions.go
@@ -3600,7 +3600,11 @@ func actionUpdateSettings(pf *PanelsFrame) {
 		AppConfig.UpdateInterval = comboInterval.Menu.SelectPos
 		SaveConfig()
 		dlg.Close()
-		CheckForUpdates(pf, true)
+		// Do not hold the UI mouse-dispatch loop while the manual update
+		// check waits for GitHub. The dialog is already closed, so the
+		// network request can safely continue in the background and post
+		// its result back through FrameManager when it completes.
+		go CheckForUpdates(pf, true)
 	}
 
 	vtui.FrameManager.Push(dlg)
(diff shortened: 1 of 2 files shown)


unxed:
Падает CI

PR #680, opened 2026-08-22
ci: check dependencies for known vulnerabilities

refaim:
`go vet` and `golangci-lint` reason about the code in this repository; nothing looked at what the module graph drags in. `govulncheck` does, and unlike a broad linter it reports only advisories the code actually *reaches* — so it sits at zero rather than accumulating a backlog someone has to triage.

Three reachable advisories were open when I ran it. All three are closed by the upgrades in this PR:

| | | |
|---|---|---|
| **GO-2026-6222** | `golang.org/x/image` v0.44.0 → v0.45.0 | Excessive memory allocation while decoding VP8L. Reached from `mediainfo.parseImage` via `image.DecodeConfig`, so a WebP the user merely browses to is enough to trigger it. |
| **GO-2026-6061** | `google.golang.org/grpc` v1.78.0 → v1.82.1 | xDS RBAC engine and HTTP/2 transport server, reached through `internal/netproxy` and the Google Drive backend. |
| **GO-2026-5506** | `go.opentelemetry.io/otel` v1.38.0 → v1.44.0 | Baggage header extraction allocates without bound, reached from the Drive backend's `init`. |

## Why otel lands on v1.44.0

That version is squeezed from both sides, and the two obvious choices are both wrong:

- **v1.41.0** is where GO-2026-5506 is fixed, but pinning there drags grpc back down to `v1.81.0-dev` through MVS, which reopens GO-2026-6061.
- **v1.43.0** is what grpc v1.82.1 pulls in on its own, but that release carries GO-2026-5158.

v1.44.0 is the first one clear of all three. Worth knowing before anyone "simplifies" it back down.

## Scope of a single run

The scan is GOOS-dependent — build constraints decide which files enter the call graph — so one invocation does not cover a tree with 44 windows-only and 8 darwin-only files. I ran it by hand under `linux`, `darwin` and `windows`, and all three report the same empty result, which is why the step runs once rather than looping. If those ever diverge, a per-GOOS loop is the fix; it is cheap, since the step takes ~11s on CI with the build cache already warm from the lint step above it.

## One thing to expect

The tool version is pinned, but the vulnerability database is fetched at run time. A newly published advisory against a dependency will therefore turn this job red with no change on our side. That is the check working as intended rather than a flaw in it, but it does mean the occasional dependency bump will arrive as "CI went red on its own" — I put a note to that effect next to the step.

Verified green end to end on a fork run: 25/25 builds, all six test cells, the race detector, and the quality job including the new step.


unxed:
Отлично!

PR #690, opened 2026-08-22
Fix Windows right-click paste rendering

Zoinen:
## Summary\n- route Windows right-click paste into the visible f4 command line through the same edit model used by keyboard paste\n- keep direct PTY paste for hidden command lines and terminal applications\n- add regression coverage for both command-line and PTY paths\n- update vtui to v0.1.268, which contains the portable 32-bit far2l bounds fix needed by the repository's cross-arch CI\n\nFixes #361\n\n— zoin-bot


unxed:
Тесты падают.

PR #706, opened 2026-08-22
lint: clear two linters' backlogs and check them whole-tree

refaim:
golangci-lint runs with `--new-from-rev=origin/main`, so it only ever reports findings on lines a pull request touched. Everything the linters have to say about the existing tree is computed and then discarded before printing — currently around 2450 findings.

That is a fair way to adopt linters on a codebase that predates them, but as set up it has no way out. `main` is a moving target, so anything that lands becomes "old" and stops being checked forever, and the size of what is being ignored is invisible to everyone. `gosec` is the extreme: it sits in the enabled list and reports almost nothing.

This starts a way out, smallest backlog first. A linter graduates to `.golangci-strict.yml` once its backlog is genuinely zero, and from then on it is checked across the whole tree and cannot regrow. Adding one to that list means clearing its debt in the same change — the list is a promise, not a wish.

The strict list is a **subset** of the main one, not a separate set, so a local `golangci-lint run` with no arguments still covers everything. Running both costs about a second: measured on CI, the first invocation takes 29s and the second 1.1s, because it reuses the first's analysis cache.

## ineffassign — 19 findings

**One was a real bug.** A panel test computed the row its hidden KeyBar was expected on, directly below a comment saying it must stay on row 24, and never asserted it. The test could not fail. It asserts now.

Sixteen were initialisers that every branch overwrote — paddle and ball colours, rune widths in the editor and viewer, an image title. Deleted.

Two were worth more than deletion:

- `vtvibe` marshalled its chat request to JSON, then added the Gemini tool block to the request and marshalled again over the top. The first pass was not merely wasted, it encoded a request missing the tools.
- `main`'s second argument pass assigned `--sudo-dispatcher` after the early pass had already dispatched on it, with nothing left to read the value.

## unused — 64 findings, 272 lines

33 methods on types that were never once constructed, 10 dead methods on live types, 9 functions, a field, a constant, a variable.

Two hazards here rather than one. A method may be the only thing keeping its type inside an interface, so deleting it drops the type out silently and the failure surfaces somewhere else entirely. And a function may be referenced solely from a file behind a build tag this platform does not compile — with 150+ constrained files and nine GOOS values, "nothing references it" on Linux means little. The second was checked by cross-building rather than by reading.

The one that needed a real decision was `findQueryMatcher.matches`, which inverted the result for the "files not containing" search mode and looked like the only implementation of it. It is not: the streaming search inverts on its own at both of its exits, and a test covers that path. The method was a duplicate nobody called.

No `//nolint` anywhere, and nothing silenced to reach zero.

## Verification

Whole fork CI run green: 25 builds, 9 vet cells, 6 test cells, race detector, quality. `ineffassign` and `unused` both report 0 across the tree, and the strict config was checked against a deliberately planted violation to confirm it actually fails rather than passing vacuously.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -514,6 +514,19 @@ jobs:
         version: v2.13.1
         args: --new-from-rev=origin/main
 
+    # The step above only reports findings on lines a pull request touched, so
+    # everything the linters have to say about existing code stays invisible —
+    # currently around 2450 findings. Rather than leave that permanently
+    # unspoken, each linter graduates here once its backlog is actually zero,
+    # and from then on it is checked across the whole tree and cannot regrow.
+    # Adding a linter to this list means clearing its debt first, in the same
+    # change.
+    - name: Run whole-tree golangci-lint for the linters with no backlog
+      uses: golangci/golangci-lint-action@v9.3.0
+      with:
+        version: v2.13.1
+        args: --config .golangci-strict.yml
+
     # Unlike the linters above this one is not incremental: it reports only
     # vulnerabilities the code actually reaches, so the whole tree is in scope
     # and the finding count stays at zero rather than accumulating a backlog.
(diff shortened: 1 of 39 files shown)


unxed:
Шикарно, спасибо!

PR #709, opened 2026-08-22
editor: open and search a large file without reading all of it

enedzvetsky:
Opening the 8 GB test file took twenty seconds before anything appeared, and Find All on it took minutes. Both are the same mistake in two places: the editor read all of a file to answer a question about part of it.

### Opening

`newEditorView` built the line index in its constructor. That was free while a large file was an `AsyncBuffer` — `ForEachRange` gets `ErrLoading` and stops — but a mapping answers every read, so mapping local files silently turned that line into a scan of the whole file, on the UI thread, before the editor existed. It was 17.9 s of the 20. A file the indexer owns now starts with an empty index and lets the background scan fill it, with the progress, cancellation and status line the scan already had.

The scan itself then walked the mapping, one page fault at a time. On this machine that is ~440 MB/s where `pread` of a few megabytes gets 2.2 GB/s from the same file — the fault is per page and waits for its own page. The scan now reads the file directly wherever the piece table says the text is still the original's, and reads the buffer where it is not, so anything typed is still scanned as edited text.

`PieceTable.OriginalRange` is what answers that question: `View` without the bytes.

### Searching

Find All handed the matcher the whole buffer, and there are only two ways to produce one — the mapping, which faults all 8 GB into residency, or an assembled copy in the heap when the buffer has been edited, which is another 8 GB. Meanwhile the line index was scanning the same file, and a fault storm starves `pread`, so the two fought over a disk one of them was using a page at a time.

The collection now reads 4 MB windows. Consecutive windows overlap by more than the longest a match can be, and the scan never looks back before the last match it reported, so nothing is found twice or lost on a seam. Matching lines are counted in the same pass rather than a second walk over the same bytes, and the occurrences list reads the line under each row through the piece table when it paints it instead of holding the buffer for as long as it is open.

The regex and whole-word paths still assemble: their patterns have no bounded length, so a window cannot stand in for the file.

### Measured, on an 8 GB file of 114.5M lines

| | before | after |
|---|---|---|
| editor on screen | 20 s | 49 µs |
| complete line index | 18.4 s | 3.8 s |
| Find All, pattern absent | minutes | 3.8 s |
| Find All matching every line | — | 4.9 s, 2.0 GB peak |
| memory during a search | +8 GB resident | +4 MB heap |

Reading 8 GB at 2.2 GB/s is 3.7 s, so the scan and the search are within a tenth of the disk. Parallel readers were tried and are not worth it: eight `pread` workers reach 4.5 GB/s on raw bytes, but every variant that also built the 114M offsets landed at 3.7–4.3 s, a single reader included.

### What a deferred index broke, and what fixes it

Anything that reports a line number gets it from the index, so a result that arrives before the scan does used to be answered with the index's last line and a column counted from there. That showed up as every Find All row on line 1 with the whole file as its text; as a viewer opening at the top instead of at the cursor; and as a hex cursor becoming "line 0, column five million".

- A search waits for the scan to reach its match — as far as the match, not for the whole file, so a hit on line 3 of an 8 GB file does not wait for the other 8 GB.
- A position known only as a byte offset is remembered and placed by the scan when it reads past it, rather than guessed at.
- A rebuild after Undo, Redo or SetText reports whether it reached the end; on a lazily loaded file it may not, and a short index recorded as complete is believed by everything downstream.
- A scan cancelled by an edit clears the indexing flag on the UI thread; its own cleanup is fenced by `editSession` and was refused as stale, so nothing restarted it.
- The file is read only while the piece table's original buffer *is* the mapping. A codepage reload or a Replace All swaps in a buffer built in memory and leaves the mapping and the descriptor in place; reading the file at the new buffer's offsets indexes one thing's bytes at another thing's offsets, silently.
- The occurrences list arms fault recovery like the editor's own paint, since it now reads the mapping on every frame: a file truncated while the list is open was a SIGBUS.

### Not in this change

Plain search (F7) still assembles the whole buffer, and word wrap on a file this size still costs a per-line array. Both are the same shape as what is fixed here and are better off as their own changes.


unxed:
> the editor read all of a file to answer a question about part of it

Изо всех сил пытался этого избегать, но всё же не вышло. Отличная работа, спасибо!

unxed:
Ой, ну как же так

```
Error: cmd/f4/editor_view.go:3475:9: ev.noteIndexProgress undefined (type *EditorView has no field or method noteIndexProgress)
Error: Process completed with exit code 1.
```

Откатываю пока

PR #715, opened 2026-08-22
lint: apply staticcheck's mechanical rewrites

refaim:
staticcheck reports 319 findings on this tree, and they are two different things: roughly 270 mechanical rewrites and roughly 50 genuine bugs. This is the mechanical half, split off so the bugs can be read on their own rather than buried in three hundred lines of style churn. A follow-up takes the `SA` findings.

| | |
|---|---|
| `QF1012` | `fmt.Fprintf` instead of `WriteString(fmt.Sprintf(...))` |
| `QF1003` | `if`/`else if` chains that are switches |
| `QF1008` | embedded fields named redundantly in selectors |
| `QF1001` | conditions De Morgan simplifies |
| `QF1011` | redundant types in declarations |
| `ST1005` | capitalised error strings |
| `S1002`, `S1021`, `S1031`, `S1040`, `ST1023` | the rest |

Applied with `golangci-lint --fix` and then read rather than trusted. `QF1001` got the closest reading, since inverting a condition is only equivalent when the operands have no side effects and short-circuit order does not matter; each was checked individually. The fixer skipped four files whose suggestions overlapped and those were done by hand, and one `S1021` rewrite exposed a `QF1011` that had not been reachable before it.

## Two ST1005 left deliberately

Both are Far Manager import messages in `plugins/envman` that go straight to a dialog through `showFar3ImportError`. They are interface copy rather than error strings, and lowercasing them would change what the user reads.

## Four errcheck findings came out of this

Worth spelling out, because it is what makes a "mechanical" change not quite mechanical.

`fmt.Fprintf` returns an error where `WriteString`'s result was already being dropped, so `QF1012` turned two silent drops into reported ones. And `QF1008` rewrote `c.Conn.SetReadDeadline` to `c.SetReadDeadline`, which made a line errcheck had never looked at into a line it had — the error there had gone unchecked since the code was written.

All four are explicit now, each with a note on why the error carries nothing to act on: a terminal that cannot be written to is reported by the next read, and a deadline that cannot be set only fails on a connection whose following `Read` or `Write` will fail anyway.

## Verification

No behaviour changes and no `//nolint`. The `SA` findings are untouched and their counts are identical before and after, check by check. Whole fork CI run green.

One unrelated thing surfaced while verifying: `TestActionUpdateSettings_ManualCheckDoesNotBlockMouseDispatch` failed once on `windows/arm64` and passed on a rerun. It also failed earlier today on a branch that touched nothing but `build.yml`, so it is a pre-existing flake rather than anything here — noting it because it looks like the same background-work-outliving-its-test class that a separate change is currently fixing.


unxed:
Супер! Конфликт надо поправить только

Issue #264, opened 2026-08-03
Автоапдейт неправильно учитывает версию при ручной сборке


montoner0:
Если собрать f4 вручную, то автоапдейт считает эту версию старее, чем есть в релизах. 
Логика вообще странно работает: если отказаться от обновления, то при следующей ручной попытке обновиться выводится сообщение, что у нас уже самая последняя версия.

Zoinen:
I am zoin-bot, running on Windows 10 x64. I am starting work on this issue. — zoin-bot

Zoinen:
I am zoin-bot, running on Windows 10 x64. Work on this issue is complete. Manual VCS builds now compare their build timestamp with the release publication timestamp, avoiding false downgrade prompts while still offering newer releases; exact release and installed-version behavior is preserved. Pull request: https://github.com/unxed/f4/pull/734. Validation: full go test ./... -count=1, native Windows x64 build/version check, Linux cross-build, and all 49 GitHub Actions checks passed; the PR is clean and mergeable with no conflicts. — zoin-bot

Zoinen:
I am zoin-bot, running on Windows 10 x64. Work on this issue is complete: manual VCS builds now use their build timestamp when deciding whether a stable release is newer, while exact release tags and the last installed version remain protected from update prompts. Pull request #734 contains the fix. Full tests, native Windows build, Linux cross-build, and all 49 CI checks passed; the PR is clean, mergeable, and conflict-free. — zoin-bot

Zoinen:
Update from zoin-bot after the maintainer merge.

The updater fix was rebased onto current `main`, the `updater_test.go` conflict was resolved while retaining both regression tests, and PR https://github.com/unxed/f4/pull/734 was merged as `c10c3c8b`. The updater-focused tests and native Linux ARM64 build/version smoke pass locally.

The refreshed post-merge matrix exposed an unrelated upstream test compile regression introduced by newer `arkanoid_test.go` changes: `rand.New` was used without importing `math/rand`. I isolated that one-line fix in https://github.com/unxed/f4/pull/743; its focused Arkanoid tests pass locally and the full CI matrix is queued.

— zoin-bot

unxed:
Как сейчас?

unxed:
Закрываем по правилу двух недель тишины.

PR #734, opened 2026-08-22
Avoid false updater prompts for newer manual builds

Zoinen:
## Summary`n- compare manual VCS builds with release publication time instead of comparing a commit hash to a release tag`n- preserve exact release-tag and installed-version behavior`n- add regression coverage and a three-pass solution review`n`n## Validation`n- go test ./... -count=1`n- native Windows x64 build and version check`n- Linux cross-build`n- go vet local output contains only existing unsafe-pointer warnings`n`nFixes #264.`n`n— zoin-bot

### DIFF
--- a/ISSUE_264_SOLUTION_REVIEW.md
+++ b/ISSUE_264_SOLUTION_REVIEW.md
@@ -1,0 +1,35 @@
+# Issue #264 solution review
+
+The report combines two symptoms: manual builds can be compared as if their
+commit hash were a release version, and declining an update could suppress a
+later manual check. The second behavior is already covered by the merged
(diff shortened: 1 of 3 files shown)


unxed:
Требуется разрешение конфликтов

Zoinen:
zoin-bot rebased this PR onto current upstream/main and resolved the only conflict in `cmd/f4/updater_test.go` by retaining both the helper-argument regression and the manual-build timestamp regression. The PR is clean/mergeable again; the refreshed CI is running. The local updater-focused tests pass after accounting for the unrelated upstream `arkanoid_test.go` missing `math/rand` import.

PR #752, opened 2026-08-23
fix: order editor teardown and quick-view refreshes against their goroutines

refaim:
Three tests reported races under -race, each reproducible on its own without
any test-order shuffling. Two of the three turned out to be bugs in shipped
code, not in the tests.

EditorView.Close tore down the buffers while an indexing goroutine was still
reading them. Close already cancelled the indexer, but cancellation only asks:
the goroutine can still be inside guardMapping when Close moves on to close the
mapping the piece table's windows point into. The comment above that teardown
says the background work has to have been "told to stop" first, which is
exactly the gap -- told, not waited for. StartIndexing now registers its
goroutines and Close joins them after cancelling.

startDirScan called FrameManager.HardRefresh straight from its scan goroutines,
and two scans overlapping meant two goroutines mutating frame-manager state at
once. Both refreshes now go through the UI task queue, which is what everything
else on that path already does.

The third, TestActionDelete_RetrySuccess, was a genuine test bug: it read the
mock's deleted list and ran its cleanup while the delete worker was still
appending. It now waits for the worker's completion handoff first.

Verified with -race -count=10 on each of the three, and the package suite.

### DIFF
--- a/cmd/f4/actions_test.go
+++ b/cmd/f4/actions_test.go
@@ -382,7 +382,7 @@ func TestActionDelete_RetrySuccess(t *testing.T) {
 	}
 
 	pf := NewPanelsFrame()
-	defer pf.Close()
+	t.Cleanup(pf.Close)
 	pf.ResizeConsole(80, 25)
 	fsp := pf.panels[0].(*FileSystemPanel)
 	fsp.vfs = mv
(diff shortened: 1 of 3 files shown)


refaim:
Детектор гонок кстати сейчас не работает нормально, занимаюсь его починкой

PR #755, opened 2026-08-23
test: check the errors these tests were throwing away

refaim:
errcheck's whole-tree backlog is 1432, and 900 of those sit in test files rather
than in shipped code. This clears the test half everywhere except cmd/f4, which
is 502 findings on its own and is where everything else in this repository is
landing right now, so it goes separately rather than spending its life in
conflict. Production code is untouched here and still has 532 findings; that is
its own pass, and the interesting ones are in it.

The default treatment is to check the error and fail. A test whose setup
silently failed does not report the setup: it reports whichever assertion falls
over later, somewhere unrelated, and whoever reads the failure starts in the
wrong place.

Three exceptions, and only these three.

A handle opened for reading gets `_ =` on Close. A handle opened for writing
does not, because there the Close error is the only word on whether what was
written actually reached the disk. errcheck cannot tell those apart, which is
why this had to be read rather than rewritten in bulk.

Cleanup whose failure genuinely says nothing -- response bodies, connections,
listeners, an os.Remove of a path t.TempDir will take anyway -- gets `_ =` and a
clause saying why.

Goroutines other than the test's own report through t.Errorf, because t.Fatal is
only valid on the test goroutine and quietly does the wrong thing elsewhere.

Lines whose whole point is that the call fails were left asserting on the error.
Turning those into t.Fatal would have inverted the test.

Two things came out of checking rather than discarding, both in the plugin half.
The mock FISH+ peer could outlive the test that started it; its cleanup now
joins that goroutine, which is a leak the discarded errors were hiding rather
than something the change introduced. And closing the reconnect handles in
checked form initially reordered them, which that path depends on, so the
deferred closes keep the original order.

Nothing else started failing, so outside those two there was no silently failed
setup propping a test up.

### DIFF
--- a/fusefs/bridge_test.go
+++ b/fusefs/bridge_test.go
@@ -269,7 +269,7 @@ func TestReadRandomAccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("open: %v", err)
 	}
-	defer h.release()
+	defer func() { _ = h.release() }()
 
 	buf := make([]byte, 3)
 	n, err := h.readAt(context.Background(), buf, 2)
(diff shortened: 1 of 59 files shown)


unxed:
Тест падает

PR #756, opened 2026-08-23
lint: use fmt.Fprintf for the sixel capability replies

refaim:
staticcheck runs over the whole tree now, so these three landed as a red main
rather than as a note on a diff. QF1012: Write([]byte(fmt.Sprintf(...))) is
fmt.Fprintf with extra steps.

The write errors are discarded explicitly rather than by dropping the return
values on the floor. These are capability replies going back to the child over
the pty, the parser has nowhere to return an error to, and a pty that has gone
away is noticed by the read side. Same reasoning as the existing unchecked
write further up the file, which is correct for the same reason even if it does
not say so.

### DIFF
--- a/cmd/f4/ansi_parser.go
+++ b/cmd/f4/ansi_parser.go
@@ -1019,11 +1019,12 @@ func (p *AnsiParser) handleGraphicsAttributes(args []int) {
 		action = args[1]
 	}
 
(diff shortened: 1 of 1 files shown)


unxed:
Отлично, спасибо!

PR #819, opened 2026-08-25
grabber: select screen area with the mouse

Zeroes1:
The feature is completely consistent with the behavior in FAR3

## Summary
Add mouse support to the Screen Grabber (Alt+Ins). The selection can now
be made with the pointer instead of keyboard-only navigation.

Behavior:
- Left button press anchors the selection.
- Dragging with the button held extends the rectangular area.
- Releasing the left button freezes the highlighted rectangle; it no
  longer changes on later pointer movement.
- A new left button press starts a fresh selection.
- Copying still happens via Enter / Ctrl+Ins, matching keyboard use.

## Implementation
- `GrabberFrame` gains a `mouseSelecting` flag and a reworked
  `ProcessMouse`: motion extends while the button is held, and any event
  without the left button (release or hover) stops the selection. The
  release test covers both `KeyDown==false` and `ButtonState==0` so it
  works across backends that report the release differently.
- Coordinates use absolute `MouseX/MouseY`, which map 1:1 to the full-
  screen snapshot.

## Notes
Platform coverage follows the existing mouse pipeline: works on Linux
(X11/Wayland), Windows GUI/gogpu, and ebiten. The legacy Windows console
backend does not deliver mouse events at all, so this (like every other
mouse feature in f4) is unavailable there.

## Test
`TestGrabber_MouseDragSelectsArea` verifies anchor/drag/release and that

### DIFF
--- a/cmd/f4/grabber.go
+++ b/cmd/f4/grabber.go
@@ -37,6 +37,11 @@ type GrabberFrame struct {
 	shiftHeld bool
 	ctrlHeld  bool
 	altHeld   bool
+
+	// mouseSelecting is true while the left button is held after a
+	// press inside the grabber, so motion events extend the
+	// rectangular selection until the button is released.
+	mouseSelecting bool
 }
 
 // NewGrabberFrame constructs an empty grabber. The screen snapshot is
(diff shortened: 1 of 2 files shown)


Zeroes1:
В обычной консоли тоже работает, что- то ии ложь написала. 🤖

unxed:
Да уж. С чего бы это обычной консоли не доставлять события мышки? В классическом Фаре всё работало же

PR #831, opened 2026-08-25
Fix bookmarks and associations saving outside Profile in portable mode

Zeroes1:
Summary
In portable mode (UseSystemProfiles=0) f4 keeps all config under <exeDir>/Profile, but BookmarksFilePath() and defaultAssociationsFilePath() built their paths directly from os.UserConfigDir(), bypassing the portable profile. As a result bookmarks.ini and associations.ini were read/written to the system %AppData%/f4/settings even in portable mode.

Changes
cmd/f4/bookmarks.go: route BookmarksFilePath() through portable-aware resolution (<exeDir>/Profile/settings when portable, legacy %AppData%/f4/settings otherwise).
cmd/f4/file_associations.go: same fix for defaultAssociationsFilePath().
cmd/f4/config.go: getConfigIniPaths() no longer reads machine-wide ProgramData\f4\settings.ini / /etc/f4/settings.ini in portable mode, so system settings can't leak into the isolated portable profile.
Non-portable behavior and paths are unchanged (backward compatible); tests pass.

### DIFF
--- a/cmd/f4/bookmarks.go
+++ b/cmd/f4/bookmarks.go
@@ -36,6 +36,13 @@ type BookmarkSet [10]Bookmark
 // ~/.config/far2l/settings/bookmarks.ini and the f4 directory without
 // renaming.
 func BookmarksFilePath() string {
+	// In portable mode (UseSystemProfiles=0) write under <exeDir>/Profile;
+	// otherwise use the system %AppData%/f4/settings path as before. Read
+	// userConfigDir live in non-portable mode so tests overriding the seam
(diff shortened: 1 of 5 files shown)


Zeroes1:
@unxed посмотри плиз...

unxed:
Посмотрю, но попозже — сейчас голова поведением встроенного терминала на Винде занята плотно :)

unxed:
Посмотрел. Всё ок, только добавь плз тесты для portable/non-portable путей.

PR #848, opened 2026-08-27
fix(colors): make exported farcolors.ini a Custom scheme

### LINKED ISSUE #845: Баг в Appearance settings
Если экспортировать настройки кнопкой Export Scheme в файл farcolors.ini, то выбор в комбобоксе Application style перестаёт менять текущую тему. Лечится ручным удалением farcolors.ini

Zoinen:
## Summary

Fixes #845.

`Export Scheme` writes a complete palette to `farcolors.ini`. Previously that file was reapplied over every built-in style, so changing Application style could not change the visible theme. The exported file is now offered as `Custom`; complete exported files are applied only for `Custom`, while partial hand-written files retain the existing overlay behavior. The export also records its base style for missing future entries and refreshes the open selector immediately.

## Validation

- `go test -p=2 -timeout 20m ./cmd/f4`
- `go test -race -p=2 ./cmd/f4 -run TestApplyColorStyle -count=1`
- `go vet ./...`
- Native ARM64 build and `--version`
- Live Linux aarch64 ANSI TTY run: exported `Custom`, then switched to `Modern` and `Classic` with `farcolors.ini` present; both changed the rendered palette.

— zoin-bot

### DIFF
--- a/cmd/f4/actions.go
+++ b/cmd/f4/actions.go
@@ -4281,6 +4281,20 @@ func actionAppearanceSettings(pf *PanelsFrame) {
 		if err != nil {
 			vtui.ShowMessageOn(dlg, " Error ", fmt.Sprintf("Failed to export colors:\n%v", err), []string{"&Ok"})
 		} else {
+			customIdx := -1
+			for i, item := range comboStyle.Menu.Items {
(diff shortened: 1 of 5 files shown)


unxed:
Падает тест

PR #876, opened 2026-08-28
cmd: lower prompt settle/recheck delays to speed up return to panels

Zeroes1:
## Summary
- Reduce `cmdPromptSettleDelay` (150ms -> 50ms) and `cmdPromptRecheckDelay` (250ms -> 100ms) in `cmdShellSession` (cmd/f4/cmd_session.go).
- Removes the ~250ms lag felt when returning to panels after external commands and batch files finish in the local Windows cmd session.

## Why it's safe
The `settle` check still requires the screen to be **unchanged across two looks** (`current == previous`) before releasing control, so a tighter window only confirms prompt stability a few frames sooner. The flickering-prompt guard (`cmdPromptMaxAttempts`, ~5s) is unchanged.

## Verification
Local build tested: return to panels after external commands / `.bat` files is now ~50-150ms instead of ~250-400ms, and the prompt does not release early on slow ConPTY scenarios.

🤖 Generated with [opencode](https://opencode.ai)

### DIFF
--- a/cmd/f4/cmd_session.go
+++ b/cmd/f4/cmd_session.go
@@ -64,11 +64,19 @@ type cmdShellSession struct {
 // cmdPromptSettleDelay is how long after a prompt mark the screen is first
 // examined. ConPTY renders the text that follows an echoed batch prompt within
 // a frame or two; a prompt still alone after this is a prompt.
-var cmdPromptSettleDelay = 150 * time.Millisecond
+//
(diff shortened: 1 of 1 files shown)


Zeroes1:
Причина предложений изменений: текущая задержка вызывает дискомфорт, c новой полегче, но не идеал тоже :)

PR #1118, opened 2026-09-10
test(vfs): cover Windows disk helpers

unxed:
## Summary

Add Windows-only tests for vfs/disks_windows.go:
- normalize and preserve Windows physical-device paths;
- probe a seekable file size while restoring its offset;
- reject embedded-NUL paths before native I/O;
- stop device enumeration when the context is already canceled.

The tests never enumerate or open a physical disk. The required status is documented in docs/LUNOBOT/coverage-disks-windows.md.

## Validation

GitHub Actions only, including the Windows test workflow.

### DIFF
--- a/docs/LUNOBOT/coverage-disks-windows.md
+++ b/docs/LUNOBOT/coverage-disks-windows.md
@@ -1,0 +1,11 @@
+# Coverage: vfs/disks_windows.go
+
+Status: Windows-only tests cover the safe helper and preflight paths without opening or enumerating a physical disk.
+
+Covered behavior:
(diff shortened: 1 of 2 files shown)


unxed:
The Windows-only disk helper tests passed the full GitHub Actions workflow and are merged in `d01c662`. They verify path normalization, seek-based size probing with offset restoration, embedded-NUL rejection, and cancellation before device enumeration without touching physical disks.

- Try it!

*Lunobot-1 (node f33f64cd91460430a21da326; MSW)*

PR #1188, opened 2026-09-17
fix: restore App Paths registry key broken by refactor

Zeroes1:
## Summary

The App Paths registry key was accidentally changed from App Paths to term.App Paths during the internal/term extraction in commit 2efbf707.

The term.App Paths key does not exist in the Windows registry, so App Paths resolution silently stopped working after that refactor.

## Root cause

PR #344 introduced the correct key:
`
Software\Microsoft\Windows\CurrentVersion\App Paths
`

Commit 2efbf707 changed it to:
`
Software\Microsoft\Windows\CurrentVersion\term.App Paths
`

This was an accidental rename during the bulk refactor — the term. prefix was never intentional.

## Fix

Restore the original key. One-line change in internal/cmdline/resolve_windows.go.

## Testing

- go vet ./internal/cmdline/ passes


Zeroes1:
close https://github.com/unxed/f4/issues/1185

PR #490, opened 2026-08-15
Add panel selection keyboard shortcuts

Zoinen:
## Summary
- add `Numpad 5` as a panel-view alias for `F3`, with Num Lock on or off
- add current-extension selection and deselection on `Ctrl+Numpad +/-` and main-keyboard `Ctrl+=/-`
- treat directories as one selection group when the cursor is on a folder
- add main-keyboard aliases for mask dialogs (`Ctrl+Shift+=/-`) and selection inversion (`Alt+=`)
- keep the primary declared shortcut stable in menus when an action has aliases

## Tests
- focused panel selection, key normalization, hotkey ordering, mask-dialog and Ctrl+Clear regression tests
- canonical Windows console build completed and launched successfully

PR #1204, opened 2026-09-18
fix(panel): restore Far selection shortcuts and numpad-free aliases

Zoinen:
Restores the panel selection shortcuts removed by #497 (reverting #490). Ctrl+Numpad +/- and Ctrl+=/- now select/deselect files with the current extension again; folders, NoExtension entries, selection restore and autofilter-hidden rows keep their intended behavior.

Ctrl+Gray +/- is standard Far Manager behavior, documented in the [official panel commands](https://github.com/FarGroup/FarManager/blob/master/far/FarEng.hlf.m4#L351-L367). The main-keyboard Ctrl+=/- aliases extend that command to laptops and compact keyboards without a numeric keypad. Code comments, KEYMAP documentation and EN/RU help explain why both forms should remain supported.

This also restores the original PR's related shortcuts: Ctrl+Shift+=/- for mask selection/deselection, Alt+= for inversion and Numpad 5 for viewing. Default menu hints retain the declared primary shortcut. OEM aliases use the current VK-based hotkey names, preserving the newer punctuation normalization instead of restoring the obsolete codec wholesale.

History: unxed reverted the original feature in dcb0bc9d on 2026-08-15 at 19:15 +02:00, citing TestArchiveVFS_Open_SeekingProgress before the beta and requesting a fixed resubmission. That archive test now waits for progress events (14700217), and the current archive suite passes.

Validation:
- Reproduced missing bindings with the real application key dispatch before the fix; the same regression passes afterward.
- Shuffled panel, keymap, application, architecture/localization audit and archive suites passed with CGO_ENABLED=0 and the system Go build cache.
- Covered keypad/OEM input with either Ctrl key, present/absent Char, kitty protocol, case folding, folders, NoExtension, unrelated marks, Ctrl+M and autofilter.
- Incremental lint passed; language formatting checked. GitHub CI and Codecov passed for 8df33134: all 25 executed checks succeeded (4 workflow-defined skips), including all desktop builds, legacy Windows/Termux builds, native Linux/macOS/Windows tests, race detection, both lint shards and Quality/Vet. GitHub reports CLEAN/MERGEABLE.

Separate branch from current upstream/main; independent of #1203.


Zoinen:
Compatibility rationale: selecting and deselecting the current extension with Ctrl+Gray + / Ctrl+Gray - is standard Far Manager behavior ([official help](https://github.com/FarGroup/FarManager/blob/master/far/FarEng.hlf.m4#L351-L367)), so f4 should support it. Ctrl+= and Ctrl+- expose the same command on the main keyboard for laptops and other keyboards without a numpad; they are accessibility/convenience aliases, not replacements for the Far shortcuts.

This is the resubmission requested after #490 was reverted in #497. The earlier rollback cited an archive progress test before the beta, rather than an objection to the shortcuts. That test passes on current upstream, and this PR adds regression coverage through the actual keyboard dispatch path so removing either form is caught.

PR #1205, opened 2026-09-18
perf(editor): stop laying out whole long lines, and add a stall watchdog

enedzvetsky:
## The bug

Scrolling a 22 MB log (92,596 lines, 870 of them over 1 KB, longest 47 KB) with word wrap off stuttered every few seconds.

With wrapping off a fragment is the entire logical line, so every frame turned **all** of it into grapheme clusters and `CharInfo` cells — 47,000 of them for one of that log's lines — to draw the ~200 columns that fit. The screen buffer then clipped the rest. The cost was per *visible row*, so a bigger terminal multiplied it.

Two changes in the render path:

- cell building stops at the right edge of the viewport;
- the text is clipped to a prefix that provably covers the visible columns before it is segmented. It falls back to the whole line where it cannot — a line holding any right-to-left run is laid out whole, because the bidi reordering of a prefix is not the prefix of the reordering.

`internal/viewer` already reads only `width*4` bytes per row; this is the editor doing the same thing.

**Measured on that log**, 200×60 viewport, Colorer on, wrap off:

| | before | after |
|---|---|---|
| worst frame | 58.8 ms | **17.5 ms** |
| mean frame | 5.8 ms | 4.5 ms |
| frames > 25 ms | 11 | **0** |

Every spike lined up with one of that file's long lines, and they are all gone.

## The editor opened black

Found on the way, and unrelated to the above. The whole text area was painted black on black from the moment a file opened until the Colorer worker's first round trip landed — a visible flash on every F4 with `Highlighter = Colorer`.

`colorerTypeSettings` used `-1` to mean "unset" for `default-fore`/`default-back`, so its **zero value meant RGB 000000 on 000000** — and zero is exactly what a highlighter holds until a session has been read. Explicit `foreSet`/`backSet` flags make the zero value mean "nothing set", which also covers `startWindowColorer`'s three failure returns.

## The instrument that found the real cause

A CPU profile answers "where did the cycles go", which is the wrong question for a freeze: a thread waiting on a lock, a page fault or a console write burns no cycles and leaves no samples, and one stalled second is lost among two minutes of ordinary frames. Two switches:

- `--stall-watchdog [d]` — writes every goroutine's stack when a unit of UI work overruns, **or when a gap interrupts a busy stretch**. The second case is the shape a freeze in input delivery takes, where nothing runs at all and there is nothing to catch. Everything goes to one log file that exists from the moment it arms, so an empty folder means it never armed rather than that it saw nothing.
- `--trace [file]` — a runtime execution trace beside it, for GC pauses and blocking syscalls.

`internal/stallwatch` is a layer-0 leaf (registered in the architecture auditor). `internal/app/framewatch_vtui.go` is inert under the default build tag: it wants a `vtui.FrameWatch` hook that is newer than the pinned vtui, and only a diagnostic build (`-tags vtuiframewatch` with a `replace`) uses it. **Happy to drop that file** if you would rather it landed after the vtui side.

## What it found

On the machine that was stuttering, the watchdog reported **no slow frame at all** — not one, across eight dumps. The UI thread was parked in its select and the input reader blocked on the console input handle, for up to **824 ms after 832 units of steady work**. The keystrokes were not reaching f4.

That is still being chased and none of the fixes here address it. They are all real and measured, but the reported stutter has a different cause.

## Dropped after review

Two changes were in the first push and have been taken back out. Neither addressed the reported stutter, and both changed behaviour in ways that want more thought than a patch under review:

- **Extending the 64 KB layout cap from the unindexed tail to every line.** `GetLineLength` stays uncapped, so on a 200 KB minified line End puts `CursorPos` at 200000 while `LogicalToVisual` snaps the caret to the cap — typing would land somewhere neither visible nor where the caret sits. It only ever served a multi-megabyte line I constructed myself.
- **Slicing the word-wrap row count** (a real 1.8 s stall on a 600k-line file). Several callers take `GetTotalVisualRows` as a hard clamp rather than a hint — `scrollViewBy`, PgDn, multicursor — so while the estimate catches up after an edit near the top of a large wrapped file, wheel scrolling and multi-caret Down near the bottom silently refuse to move.

Both are worth doing properly; neither belongs in this PR.

## Testing

Regression tests for each change that stayed: the zero-value attribute, the viewport-bounded cell build, the clip (including that an RTL line is left whole, and that an Indic line still fills the viewport — that one fails at 131 of 200 columns without the fix), and the watchdog's two shapes of freeze. `internal/editor`, `internal/textlayout`, `internal/viewer`, `internal/terminal`, `internal/stallwatch` and `cmd/f4` (the module-wide auditors) all pass, and `stallwatch` passes under `-count=4`.


enedzvetsky:
Reopened with the coverage addressed. Patch coverage was 65%, and `internal/app/bootstrap.go` was 38 of the 71 uncovered lines on its own.

The cause was structural rather than missing tests: the two new switch cases and the block that arms the trace and the watchdog all sat inline in `Main`, which wants a terminal and no test exercises.

Moved to `internal/app/diagnostics.go`, where the decisions are separable from the run — `traceFileArg` and `stallWatchdogLimit` read the optional values, `argAfter` does the bounds check the cases were repeating, and `armDiagnostics` starts what was asked for and returns both the shutdown function and the line the user is told. `Main` keeps the same behaviour in a third of the lines.

`bindFrameWatch` now reports whether it bound the vtui hook, and the startup notice uses that to say **what** is being watched — "the whole UI loop" for a build with the hook, "editor frames" without it. An empty log means different things in the two cases, so it is worth saying out loud. It also gives the function a statement to cover, which an empty body did not have.

New tests cover both optional-value flags (including the filename that must not be eaten), the watchdog arming and naming its log, a watchdog and a trace that cannot be written, the `MaxDumps` cap, a zero limit, the clip's tab and zero-width column counting, the doubling window on heavily decomposed text, and `default-back` without `default-fore`.

`internal/stallwatch` goes 82% → 91%; the changed functions in `internal/app/diagnostics.go` and `internal/editor` are at or near 100%, apart from error paths that need a filesystem failure mid-write.

Both build paths verified: the default build, and `-tags vtuiframewatch` against a vtui checkout with the hook.

Note the coverage report in this thread was against `f05dfd7`, the first push. `internal/textlayout/wrap.go` appears there but no longer has any change in it — the row-count slicing came back out after review.

unxed:
Линтер падает

